PR #35, opened 2014-04-09
Add remove-mismatch option on spacewalk-data-fsck and added it to the man page

tchellomello:
This pull request adds the funcionality on spacewalk-data-fsck to remove the RPM which does not match checksum. 

The second commit adds on spacewalk-data-fsck man page the explanation about the new option. 

Note: this patch have the suggestion proposed by skoty-tkasparek at https://github.com/spacewalkproject/spacewalk/pull/32

Thanks.

### DIFF
--- a/backend/satellite_tools/spacewalk-data-fsck
+++ b/backend/satellite_tools/spacewalk-data-fsck
@@ -211,7 +211,7 @@ def restore_file_path(file, row):
     log(0, "File %s%s successfully restored, new file path: %s" % (prefix, row['path'], new_path))
     return 0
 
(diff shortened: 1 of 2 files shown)


tlestach:
I applied the patch, incorporated Michael's comments and did some extra changes on top. I committed the result as: 6ee97c5fe581ed9d5b1b26e95b668c7e816e1591. a9b2b7a367e04074a122043087980da75f321ba4 accepted as is.
Thank you for the idea and contribution, Marcelo! And also thanks to Michael for valid comments!

PR #112, opened 2014-07-01
Implement the new rpm weak dependency tags

dmacvicar:
Both the old IDs and the new ones are read in sequence. That is why
the self[tag] array is initialized only once and not on each _populateTag call.


sdherr:
Committing to Spacewalk master, thanks!
c27cd84acfe1b095b1996d401d3dfa23ee620d0c

PR #128, opened 2014-07-24
1122974 - ISE in case no system is selected for deployed configuration files comparison.

dyordano:
1122974 - ISE in case no system is selected for deployed configuration files comparison.


ggainey:
Incorrectly closed - 982e636 is the commit for PR#173. Reopening so it can be *actually* fixed.

ggainey:
Accepted as 70291fb1 - thanks for the patch!

PR #134, opened 2014-08-07
1125982 - "Configuration" tab should be available for an organization only in case of Provisioning entitlements are provided.

dyordano:
1125982 - "Configuration" tab should be available for an organization only in case of Provisioning entitlements are provided.

### DIFF
--- a/schema/spacewalk/postgres/packages/rhn_entitlements.pkb
+++ b/schema/spacewalk/postgres/packages/rhn_entitlements.pkb
@@ -1114,6 +1114,12 @@ as $$
                                          group_type,
                                          new_ent_count);
 
+        if group_label_in = 'provisioning_entitled' then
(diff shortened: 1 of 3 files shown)


ggainey:
The problem with hiding access to cfg-mgt once you've had it, is that you lose access to the cfg-files you have entered into SW, and you have no way to *unsubscribe* systems from cfg-mgt channels.

To really respond to losing-all-cfg-files, we would need to unsubscribe all systems subscribed to cfg-management, while still letting the admin have access to the cfg-files/revisions they've uploaded into SW.

I think the potential impact to the customer outweighs the 'correctness' of hiding the pages when you lose access. Currently leaning towards rejecting the change, but open for discussion.

jdobes:
No update for several months, closing. Please reopen if status changes.

PR #145, opened 2014-09-03
support while loop syntax for Cheetah templates

swarmingbee:


xkollar:
You must have heard that we like patches when you put patch into patch so we can apply why we apply :-). Looks reasonable. Rebased as ab477e498a4aef8ee5bcdc4271679109742ff8a9. Thank you for your contribution.

jdobes:
two more related commits:
728b822d121b9d821775c643b1978839c52c8cc9
591035f381518cdc2f76b6ffe165e519ac282f58

PR #165, opened 2014-10-24
Spacewalk proxy changes

mcalmer:
align rhn-proxy script to spacewalk-servicescript and make it systemd aware

### DIFF
--- a/proxy/proxy/rhn-proxy
+++ b/proxy/proxy/rhn-proxy
@@ -10,7 +10,7 @@ forward_services() {
     ACTION="$1"
 
     for service in $SERVICES; do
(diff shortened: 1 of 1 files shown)


sdherr:
Hi mcalmer,
Looks good, thanks! The only modification I'll make is to leave the condrestart command in the script just as an alias to restart for backwards-compatibility reasons. The restart command ignores the non-zero return code from stop if the services were not running, so they should be functionally equivalent anyway.

Committing as 28a9064c3e6a4fe7f68f2af195405847602a98d5 and 9d7afd34de4d435a9f6eadb46403992e88b4e567, and my modification as 5fc4d0e797dca40aafac8b7572b68becd608943f.

PR #168, opened 2014-10-28
spacecmd: fix listupgrades

flavio:
Do not rely on spacecmd's internal cache to rebuild the full package name when listing system upgrades. Java's `system.listLatestUpgradablePackages` API returns all the information we need since commit 3f6829d6aa6 made on the beginning of May 2010.

Relying on the cache caused some users to have 'None' displayed instead of the package name when the cache went out of sync.


tlestach:
Rebased as be7f3d9026c00517dcfd06a096e29dec0b647fba. Thank you for contribution!

PR #175, opened 2014-11-17
Fix rhn channel newest package after cloning

mcalmer:
fix rhnChannelNewestPackage table by using refresh_newest_package function again

How to reproduce:
    \* clone a channel with "select errata" so that some packages are missing in the cloned channel.

```
* Now compare packages of these two channels.

Result => no difference, because the compare query uses rhnChannelNewestPackage table
```

### DIFF
--- a/web/modules/rhn/RHN/DB/ChannelEditor.pm
+++ b/web/modules/rhn/RHN/DB/ChannelEditor.pm
@@ -745,32 +745,5 @@ EOQ
   return $exists;
 }
 
-sub clone_newest_package {
-  my $class = shift;
-  my %attr = validate(@_, {from_cid => 1, to_cid => 1});
-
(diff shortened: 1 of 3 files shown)


sdherr:
Hi mcalmer,

The clone_newest_package logic should still be valid if you are cloning a channel as current, but I agree that that is the only time it should be called (and specifically that it should not be used when selecting errata). The clone channel page I recently ported to java and I intend to do the errata clone page soon so I don't think this code will ever see direct use, but I'm committing it anyway as a reminder to myself to make the clone errata page call refresh_ instead of clone_.

Committing as 3dd8dcd38d3a93414d00748070eb616c4bab81f1 and 9ffd525bf0efa3182059706ba5990450896ef1de. Thanks!

PR #177, opened 2014-11-20
Bugfix: Binary upload crash and custom incoming file size calculations (bnc#901927)

isbm:
We found that in SUSE Manager ISE happens when one is trying to upload a binary file, because ``mark(int limit)`` method is not always supported (hence ``markSupported()`` boolean). To overcome this, the stream needs to be wrapped into a buffered stream, which has support.

Additionally, we found that the file size is calculated on JSP actions by their own, instead of calling ``ConfigFile.getMaxFileSize()`` and if one would like to have hard limit in this method (i.e. ``rhn.conf`` can specify anything, but limit is up to fixed top), then JSP would display wrong number.

Hope this helps.


ggainey:
Looks good to me - thanks for the fix and the cleanup!  Rebased as e06420 and c1c4a4

PR #179, opened 2014-11-21
1166651 - Software product name is not provided.

dyordano:
1166651 - Software product name is not provided.


xkollar:
What about only adding hostname? That should be sufficient to identify issue.

xkollar:
No progress for several months. In case something changes, please reopen.

PR #180, opened 2014-11-21
Remove "Add Selected to SSM" button where it makes little sense

renner:
On the system overview page as well as on SSM system listing the "Add Selected to SSM" button (at the bottom of the list) makes little sense. Simply selecting the systems in the system overview adds them to SSM anyways, while inside the SSM systems listing it obviously makes even less sense to have that button.

As far as I could see we actually want to have this button **only** on the Patches -> Affected Systems page. Somehow it now comes with every system listing though, while we need to explicitly hide it using the `noAddToSsm` variable on most of them (see 4b79a4145751a85b90715ae6a01e677ae301aa5b).

This patch removes "Add Selected to SSM" from System Overview and SSM system listing in the way introduced by the mentioned patch. A better fix IMO would be to add that `AddToSsmDecorator` only where it is needed making `noAddToSsm` obsolete.

### DIFF
--- a/java/code/webapp/WEB-INF/pages/ssm/systems/list.jsp
+++ b/java/code/webapp/WEB-INF/pages/ssm/systems/list.jsp
@@ -15,6 +15,7 @@
 <c:set var="notSelectable" value="true"/>
 <c:set var="showLastCheckin" value="true"/>
 <c:set var="noMonitoring" value="true"/>
+<c:set var="noAddToSsm" value="true"/>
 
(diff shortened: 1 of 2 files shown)


xkollar:
Your work has been rebased as
* 7f3cbcfe660d7a9b08a55fcb5789ebd4119c1858
* 64d3df3b783c96548d53f31601c0e8322c23d8bc
Thank you for your contribution :-) (This was annoying me for some time now.)

PR #183, opened 2014-11-28
Fix spacecmd schedule listing for negative deltas

Kilian-Petsch:
Negative timedeltas were detected as unknown options by the parser,
  which caused the listing to fail


ggainey:
Looks good, thanks for your contribution! Rebased as e00561

PR #184, opened 2014-12-01
Forward to a "raw mode" edit page in case of uploaded kickstart profiles

renner:
In the UI, when going via Bare Metal IP ranges to an **uploaded** kickstart profile:

    Kickstart -> Bare Metal -> [klick on IP range of *uploaded* profile]

You should go here:

    kickstart/KickstartIpRangeEditAdvanced.do

Instead you get here:

    kickstart/KickstartIpRangeEdit.do

The attached patch adds a redirect to struts-config.xml and lets the action class perform a check similar to the one in [KickstartDetailsEditAction.java, line 89] (https://github.com/spacewalkproject/spacewalk/blob/caf00fb26caee9642a554b770119d1892cfd9f4c/java/code/src/com/redhat/rhn/frontend/action/kickstart/KickstartDetailsEditAction.java#L89).

I don't like too much that comparison I am doing with the path, but it was needed for differentiation since both pages are using the same action class. Let me know if you have a better idea here, or if you'd prefer to somehow refactor action classes in order to avoid such path comparison.

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/action/kickstart/KickstartIpRangeAction.java
+++ b/java/code/src/com/redhat/rhn/frontend/action/kickstart/KickstartIpRangeAction.java
@@ -125,6 +125,13 @@ public class KickstartIpRangeAction extends RhnAction {
         }
         // display the ranges (if any), and allow user to add ranges (dynaform)
         else {
+            // Forward to the raw mode edit page in case this is an uploaded profile
+            if (ksData.isRawData() &&
+                    !mapping.getPath().equals("/kickstart/KickstartIpRangeEditAdvanced")) {
+                return getStrutsDelegate().forwardParam(mapping.findForward("raw_mode"),
+                        RequestContext.KICKSTART_ID, ksData.getId().toString());
+            }
+
             setupFormValues(form);
         }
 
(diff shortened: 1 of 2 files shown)


sdherr:
Hmm. What we would do in similar situations is to have a c:choose or similar in the jsp that draws the Bare Metal page such that it creates a link directly to kickstart/KickstartIpRangeEditAdvanced.do if the profile is uploaded. Would that be possible instead?

renner:
Thanks for the review! I will try out what you are suggesting so we can see if that works and how the patch looks like.

tlestach:
Makes sense. Rebased as 3f9ad94c36ec8bc52b5ceb111ae753613c5c1e3b. Thank you Hannes!

tlestach:
Oops, I totally missed sdherr's comments. I do not plan to revert the commit, but in case there's the discussed patch available, we're happy to review. I am sorry.

renner:
I am sorry for not having provided a solution as it has been suggested by @sdherr, I didn't find the time yet to try that out. Thanks for merging anyways, this solution is also acceptable I would say. In case I have the other patch available I will make another PR of course.

PR #185, opened 2014-12-02
fix call of setCustomOptions()

mcalmer:
setCustomOptions expects a String[] but ksdetails['custom_opts']
is a dict with keys "name" and "arguments". This results in a

ClassCastException: redstone.xmlrpc.XmlRpcStruct incompatible with java.lang.String


ggainey:
Accepted as a6f385 - but see 985030, the first custom_opts has no 'arguments' and throws an error, so I added some protection for that case.

PR #187, opened 2014-12-03
Fix various exceptions causing ISEs on GET /rhn/common/DownloadFile.do

renner:
This patch fixes various exceptions that are otherwise causing ISEs when receiving requests to `/rhn/common/DownloadFile.do` (which can happen with e.g. Qualys security scanners).


sdherr:
Looks reasonable, committing as 13b4fce67e6e68681eeb844f53c57b784a6ba4f9 through 30082ae88fadd9ea13a5bd1ec9c5662a613a1d32. The only change I'll make is that parseDistUrl needs to explicitly return null in the case of a malformed url, because handleKickstartDownload specifically checks for null and otherwise it would never be null (just an empty or partially-filled map) and I don't think we should be in the business of trying to respond to a malformed url.

PR #188, opened 2014-12-03
Remove "Select All" button from system currency report

renner:
This commit removes the "Select All" button from the system currency report. It makes sense since nothing will actually be selected when you click it. Find this page at:

https://\<hostname\>/rhn/systems/SystemCurrency.do


tlestach:
Rebased as 59873b98b9c58f0ece151d192565deabfe612289. Thank you for contribution!

PR #192, opened 2014-12-08
Make cobbler detect os version for CentOS 7

phurrelmann:
Patch cobbler20 so that it properly detects the os version for CentOS 7.
Same fix has been pushed upstream with https://github.com/cobbler/cobbler/pull/1021


sdherr:
Hi,

Looks good, committing the cobbler commit as 24082e27da1eea84a83f2321d1df9d2ec3892710. The spacewalk-setup-ipa-authentication change looks like it was already committed as 899861c0b8c0e36813bc1943787867565aad8e12. Thanks!

PR #193, opened 2014-12-08
Fix ipa configuration of tomcat-service for CentOS7.

phurrelmann:
Tomcat7 on CentOS7 uses /etc/tomcat/server.xml. Adjust regex to match.

This fixes:
Use of uninitialized value $tomcat_service in concatenation (.) or string at /bin/spacewalk-setup-ipa-authentication line 337.


tlestach:
Reabsed as 899861c0b8c0e36813bc1943787867565aad8e12. Thank you.

PR #194, opened 2014-12-09
fix configchannel export - do not create 'contents' key for directories

mcalmer:
Import a directory which has a 'contents' key failed with

InvalidArgsException: Invalid argument(s): contents


ggainey:
Rebased as 3ff691 - thanks for the fix!

PR #195, opened 2014-12-09
Set cobbler hostname variable when calling system.createSystemRecord

dmacvicar:
When adding a record for a non-registered system, there is no way to change the
hostname variable which is used in the autoinstallation profiles.
So we set it from the server hostname or name, depending which one is available.


sdherr:
Committing as a27b1622ee1a829108a8430e66e1ade73b33b311. Thanks!

PR #196, opened 2014-12-11
Apidoc fixes (regarding DocBook XML generation)

renner:
Here is some changes about generating apidocs as DocBook xml, we use that for making a PDF file. In order to produce valid XML it was necessary to adjust the xmlrpc.doc comments in SystemHandler.java a little. Apart from that there is some typos and grammar fixed also for the jsp and html verisons of the apidocs.


sdherr:
Committing as 0bc4dbb94f1a50073ed8341f03e9ee4be6ac2248 through c7bd6c073ff4b03649e8cf38665ff6cc94278265.

PR #197, opened 2014-12-11
spacecmd: handle erratas similar to packages (offer _diff and _sync functions)

joergsteffens:
and also 2 other minor improvements.


ggainey:
Thanks for the new commands!  Committed as 951c7a 37d28a 862ca9

PR #198, opened 2014-12-15
fix setting powermanagement values

mcalmer:
save when the value has changed and not when it is equal

fixes 93adb3fc2f067d7c0a2f5b2b29340311512e3dbb


sdherr:
Looks good, but I did squash things down into only two commits. Committing as 348018b843d664335b564410671118a357a30236 and f1f6d6059519d9c4aedb22cc4acdbc3cd9251c2a. Thanks!

PR #199, opened 2014-12-16
Sort filelist in configfile.compare event history alphabetically

mcalmer:
Makes it easier to find a file in the list if a lot of files
gets compared. Additionally it is less confusing if the order
do not change with every visit of the history page.


sdherr:
Committing as b69802cf2c0d2c4c52cdf89c78837b293c04e259. Thanks!

PR #200, opened 2014-12-17
Upstreaming of misc user interface polishing and fixes from SUSE Manager

dmacvicar:
This are the Spacewalk relevant fixes that we have done to SUSE Manager during the last week. We excluded changes to our theme. All the changes are semantic so they are independent of the style used.


tomas-kasparek:
Hello, I've committed your PR into Spacewalk and also applied checkstyle fix.
Thank you for contribution.

PR #201, opened 2014-12-17
Fix "Select All" buttons for the rhn:listdisplay tag

renner:
This should fix the "Select All" buttons whenever the rhn:listdisplay tag is used. In the latest versions it was never actually rendered, see e.g. in this place (make sure to have something in the list there):

/rhn/systems/details/configuration/DiffFileSubmit.do?sid=xxx

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/taglibs/ListDisplayTag.java
+++ b/java/code/src/com/redhat/rhn/frontend/taglibs/ListDisplayTag.java
@@ -484,11 +484,6 @@ public class ListDisplayTag extends ListDisplayTagBase {
         out.append(target.toString());
     }
 
-    @Override
-    protected void renderHeadExtraAddons(Writer out) throws IOException {
-        renderSetButtons(out);
-    }
-
     private void renderSetButtons(Writer out) throws IOException {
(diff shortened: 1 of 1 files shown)


xkollar:
Hello Johannes. Can you expand provided example in terms of:

* Steps to reproduce
* Observed behavior
* Expected behavior

It would certainly help the case. Thanks.

dmacvicar:
There are two lists implementations in Spacewalk. Also, there are two "Select All", the checkbox that selects the current page and the [Select All] button that selects All elements.

There are two fixes in this commit:

1) Make the [Select All] button work at all (it did not because the param name was wrong)
2) Move the [Select All] button from the top right to the bottom left like it was before before, and like it was now in the other list implementation.
It looks like this now:

screenshot from 2015-01-20 11 52 56

Before, it looked like this:

screenshot from 2015-01-20 11 59 01

Now it is consistent and works. We should still consider removing the [Select All] completely and do what GMail does when you click the checkbox. But that is another topic.

sdherr:
Committing as b7a0792732f3ad1c7aadc672e633f3ede0186b57 and ec4b8fb51bfc879c423d306bb7c553807f972172. Thanks!

PR #205, opened 2015-01-02
convert empty string to null for DMI values

mcalmer:
Prevent:
IntegrityError: new row for relation "rhnserverdmi" violates
                check constraint "vn_rhnserverdmi_product"

### DIFF
--- a/backend/server/rhnServer/server_hardware.py
+++ b/backend/server/rhnServer/server_hardware.py
@@ -816,6 +816,8 @@ class DMIInformation(Device):
         Device.__init__(self, fields, dict, mapping)
         # use our own sequence
         self.sequence = "rhn_server_dmi_id_seq"
+        self._autonull = ("vendor", "system", "product", "asset", "board",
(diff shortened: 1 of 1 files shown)


sdherr:
If null values are allowed in the table what's the point of having a "not empty string" constraint? Should we just remove the constraints from the table instead?

sdherr:
Okay, well after poking around for a few minutes I couldn't figure out where those constraints are being auto-generated, so committing this fix since it's similar to things we do other places in that file.

Committing as 2590b950597c981d03f1b0dd0c09384078894372.

MichaelMraka:
% If null values are allowed in the table what's the point of having a "not empty string" constraint?

In Oracle empty string is inserted as NULL and the life is easy. On the other hand in PostgreSQL empty string and NULL are two very different values. So to have both db backends compatible there's code in schema/spacewalk/postgres/end.sql which generates 'non-empty string' constraints for every varchar column in database. And don't even think about disabling them... never, ever... I warned you ;).

PR #206, opened 2015-01-02
Make rhn-proxy command use systemd

phurrelmann:
Make rhn-proxy command use systemd in the same way as spacewalk-service does


sdherr:
Thanks for the patch. It looks fine, unfortunately the same change was already submitted as https://github.com/spacewalkproject/spacewalk/pull/165. I can only commit one of them, so I chose that one since it was older. Thanks for the contribution though!

PR #207, opened 2015-01-05
Fix wrong package dependency using yum without priorities

lbayerlein:
It is the same as pulled a short time ago on the spacewalk-backend package. Yum would install the spacewalk-pylint package from the epel-release. There its version 0.5 and it produces errors on the package build with tito.


tlestach:
Rebased as bc0e461b515874bee693c6436bc794c78fe11b0f

PR #208, opened 2015-01-06
Make cobbler20 require syslinux

phurrelmann:
Make cobbler20 require syslinux. Upstream versions require it, too. This fixes BZ#988329


tlestach:
Rebase as da111e15523d88daa74481a93f8505437f2c68b9. Thank you!

PR #210, opened 2015-01-07
Fix 403 errors for proxy wsgi requests

phurrelmann:
Fix 403 errors for proxy wsgi requests


tlestach:
Rebased as cd295aa205874bce249b5c1ba0d8d2f4b9a10183. Thank you for the contribution!

PR #211, opened 2015-01-10
move %pre section down and eliminate an %if

mcalmer:
make the spec a bit easier to read


tlestach:
Rebased as 338dfe6b0d676e8839c882ac14557ab8b22c671d. Thank you.

PR #214, opened 2015-01-19
Tiny changes in strings on the webUI

jhutar:
Just fixed name of a button referenced in some help message and made text in a link for channel deletion look similar to rest of similar texts.


xkollar:
Hi Jan. Thanks for the pull request `:-)`. Would it be possible to update it so that all mentioned files gets updated?

``` .bash
git grep 'delete software channel'
```

jhutar:
I have did something, but please check it through-fully.

xkollar:
Rebased as 2ed8f3261cfb27bdc4317d02f67787b4989c5179. Thank you for the contribution. :+1:

PR #217, opened 2015-02-04
add ability to specify gpg bits from spacecmd when creating software chans

larkit-ian:
While the API allows one to specify GPG bits when creating channels, spacecmd currently doesn't.  This patch should allow spacecmd to do so.  Feedback welcome. ;)


ggainey:
Fixed a minor typo, but otherwise looks good - committed and tagged as spacecmd-2.3.16-1.  Thanks for the feature!

PR #219, opened 2015-02-15
Remove more monitoring

mcalmer:
remove more monitoring from spacewalk.
- config files
- setup 
- proxy
- etc.

There are still references in:
- selinux/spacewalk-monitoring-selinux/
- schema/spacewalk/common/data/rhn_config_parameter.sql
  Can this table be dropped?
- schema/spacewalk/common/data/rhn_config_macro.sql
  Can this table be dropped?
- java MonitoringConfigFactory.java
- java ConfigMacro.java

@sdherr: please review.


sdherr:
The submitted changes look reasonable, committing as 315e8f988ccb6835dcfa06920499245801c6d56b to 59b9f40b27ee5a26305984ff20d66f49cffa2c80. Thanks!

The monitoring selinux package is probably not needed any more, but I'd have to look it over carefully to be certain of that. The reason the other classes / tables are still around is that there are one or two things that are sill supported that required them. For example the ConfigureSatelliteCommand (used for /rhn/admin/config/GeneralConfig.do) depends on ConfigMacro and MonitoringConfigFactory, and the hbm mappings for ConfigMacro include those tables. I didn't want to break things during the giant monitoring removal, so I left all that stuff in. However a more careful investigation / refactoring may reveal that it's really not necessary any more.

PR #220, opened 2015-02-16
init the second DB connection only when needed

mcalmer:
__test_DB2() is called when this second connection is needed.

In case the connection is not yet open, it calles initDB(initsecond=True) and try to open it.

To correctly deal with not all connections esteblished or not, call closeDB() if initDB() thows and error.

### DIFF
--- a/backend/server/rhnSQL/__init__.py
+++ b/backend/server/rhnSQL/__init__.py
@@ -104,13 +104,16 @@ def __init__DB2(backend, host, port, username, password, database, sslmode, sslr
 
 
 def initDB(backend=None, host=None, port=None, username=None,
-           password=None, database=None, sslmode=None, sslrootcert=None):
+           password=None, database=None, sslmode=None, sslrootcert=None, initsecond=False):
     """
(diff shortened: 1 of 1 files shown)


mcalmer:
@sdherr : our testsuite looks good so far.
Maybe you can have a look if  time permits

sdherr:
I think this looks reasonable, committing as 7e5f3bb9dda342c4c08e64fa66aae56de30ccd97.

The only major problem I can foresee is that if another tool / module is calling rhnSQL.initDB() with params *and* that thing needs to use the secondary db connection, the secondary db connection will always be instantiated with the configured db settings and not with what the caller passes in, potentially breaking the tool. However calling initDB with params is very unusual, and it looks like the only things that call it with params are tests in backend/server/tests (which should be updated anyway) and the scripts/update_symlinks.py script (which doesn't need the secondary connection). If anything is written in the future that needs to call initDB with params and needs to have the secondary connection configured properly, then it will just have to know to call the method once for each connection.

PR #221, opened 2015-02-16
Fix the primary key definition for rhnPackageFile

timspeetjens:
This fixes https://bugzilla.redhat.com/989202


sdherr:
Cherry-picking as 3d9535390c3fade2f7e590df6c3d40962baca59f. Thanks!

PR #222, opened 2015-02-17
Catch NumberFormatException and send error to the client instead

renner:
This fix prevents the webapp from crashing with an ISE which in turn can lead to a mail storm when security scanners are sending fake requests to the server (e.g. Qualys). Can be reproduced using the following URL:

- https://\<spacewalk-server\>/rhn/common/DownloadFile.do?url=/a/b/c

See also the closed #187 that was about fixing similar issues.


tlestach:
Interesting.
Rebased as 3afe712e7eb6e91525393eccc81a704a16cad931

PR #223, opened 2015-02-19
Fix generation of pre-requires (pre="1") in rpmmd metadata with newer rpms

dmacvicar:
Nowadays not all pre-requires come from the RPMSENSE_PREREQ sense flag. "requires" dependencies related to rpmlibs may be represented using RPMSENSE_RPMLIB in newer distros (eg. SLES-12).

At the end all maps to pre="1", but we need to recognize all of them.


sdherr:
Committing to Spacewalk master, thanks!
37d7d168daa84f46dc63f68234570458185cce78, f820986688ae3c5cf4f793609f116ec9a2159b4b

PR #224, opened 2015-02-19
1194418 - API call channel.software.clone does not work as expected for child channels.

dyordano:
API call channel.software.clone does not takes in account the 'parent_label' parameter any longer and always the result channel is a Parent/Base channel instead of Child one.

The Regression was caused by code refactoring and in particular by
SW commit: 2984025c0b3e0c4f044eafe9f363fe1b2bd3d585
    Port Errata Clone page from perl -> java
    Make nav link to java channel clone and errata clone pages
    Also make various clone errata jsps share common list


ggainey:
Looks good - commit f55909f57b41727074061339c6bf290366da0e32, tagged as spacewalk-java-2.3.143-1

PR #225, opened 2015-02-20
BZ 1194790 regenerate errata cache asynchronously when subscribing server to channel

tpapaioa:
BZ 1194790 regenerate errata cache asynchronously when subscribing server to channel
https://bugzilla.redhat.com/show_bug.cgi?id=1194790


tomas-kasparek:
Committed into Spacewalk as e8c453fe00dc8ef56851e8dfe39cda043144a594.
Thanks for contribution.

PR #227, opened 2015-02-23
Spacewalk channel enhance

xkollar:
Having to touch the code, I decided to spent some free time on enhancements... does those make change?


ggainey:
These all look reasonable - committed as rhn-client-tools-2.3.12-1

PR #228, opened 2015-02-26
missing value PIDFile for systemd

Pajinek:
Missing value PIDFile=/var/run/rhn-search.pid

Command "systemctl status rhn-search" returns exit code 3
Active: inactive (dead) since St 2015-02-25 17:30:07 CET; 26min ago

I add PIDFile and systemctl return 1. When I run "systemctl enable rhn-search" and then I reload systemctl, the systemctl status is 0.


tlestach:

PR #229, opened 2015-02-26
Have a separate logfile for the web UI instead of logging to catalina.out

renner:
This patch makes the web app log into `/var/log/rhn/rhn_web_ui.log` instead of `catalina.out`. The `catalina.out` file is just meant to be the stdout of the tomcat process redirected and applications are actually supposed to log somewhere else. Feel free to suggest a different name or locations for the logfile if you want.

I didn't setup logrotate for this file, rather I used the RollingFileAppender with `MaxFileSize=10MB`. Further I didn't change the URL of the page where the logfile can be viewed (`/rhn/admin/Catalina.do`), let me know if you would want this to be changed.

Hope this patch is complete apart from that.


eherget:
This PR has been sitting a while and would cause some confusion by changing log location.  Closing.

PR #230, opened 2015-03-03
Check for reboot type

mcalmer:
functions named "reboot_in_progress" should check for reboot type only.

This could fix https://bugzilla.redhat.com/show_bug.cgi?id=1187358


sdherr:
I agree. Committing as fa3fae6a8f2aab2302fd31b8eb5072c5c2dad9f5 and 9316c985c7119ec93a6b44e8676bb1edaa8bd3da.

PR #231, opened 2015-03-04
Support for a threshold on the number of clients executing actions

dmacvicar:
While pull has a randomized interval, if you schedule hundred of actions via osa then you will end with hundred of clients going back to the server at the same time.

This threshold parameter allow to plan for a number of slots according to your server capacity.

Before trying it out, make sure you have 6188430e5c0a4b60c6ac42547934fc3035581c9f
from https://github.com/spacewalkproject/spacewalk/pull/230 which fixes another bug.


sdherr:
Looks great, thanks! Committing as cb29de1b9a33be184c652d375530f4feaeb7e1d1 through 3daa3ca8f18c86579689a318dc59721f2a816cc1.

PR #232, opened 2015-03-04
osad to execute rhn_check without blocking the callback

dmacvicar:
osad does a fork and waitpid() inside the _message_callback, which can block for long time. During that time the server continues to notify and those messages get accumulated. Once rhn_check returns, all messages are processed without any context and tons of rhn_check calls are fired back to the server.
Also I observed time ago that blocking for a very long task timeouts the connection.

This changes the way rhn_check is executed. We use a subprocess handle to have always a handle to the process. We don't block and can react to _message_callback while the process is running.
When rhn_check is running, we change the select() timeout from 180 to 5 in order to monitor the process.

If the process fails (rhn_check crashes), then in case the server is not talking to us, _process_hook knows that we are in failed state and forces rhn_check in order to exhaust the remaining_tries or hope that it executes correctly, so that the action gets failed at the server side or completed.

Once the rhn_check finishes correctly, we go back to the 180s poll interval. Whether to use 5 also for the case of (no rhn_check running, last one failed) can be discussed. For (rhn_check running) case we definitely need a shorter interval to monitor the process.

How to test this?

I sent 3 one after another remote execution scripts with "sleep 10", and then a fourth one with "sleep 10; killall rhn_check"

When the process is started, you should see it,a nd also you should see that it changes the poll interval to monitor the process:

```
2015-03-04 09:09:08 osad_client.run_rhn_check_async: About to execute: ['/usr/sbin/rhn_check']
2015-03-04 09:09:08 osad_client.run_rhn_check_async: executed /usr/sbin/rhn_check with pid 16469
2015-03-04 09:09:16 jabber_lib.process: 5
```

When more messages come, while rhn_check is running, the message is still processed:

```
2015-03-04 09:09:16 osad_client.run_rhn_check_async: rhn_check is still running, ignoring notification...
```

When rhn_check is killed, it will realize it and rhn_check will be called either at the next message callback or in the process_hook even if the server does not message us.

```
2015-03-04 09:10:04 osad_client.process_loop_hook: rhn_check exited with status -15
```

```
2015-03-04 09:10:04 osad_client.run_rhn_check_async: rhn_check failed last time (fail count 2)
2015-03-04 09:10:04 osad_client.run_rhn_check_async: About to execute: ['/usr/sbin/rhn_check']
```

After about 9 retries the action would fail at the server and osad will return to normal and to the 180 interval.

### DIFF
--- a/client/tools/osad/src/jabber_lib.py
+++ b/client/tools/osad/src/jabber_lib.py
@@ -525,7 +525,7 @@ class JabberQualifiedError(JabberError):
 
     __str__ = __repr__
 
-class JabberClient(jabber.Client):
+class JabberClient(jabber.Client, object):
     _seq = 0
     BLOCK_SIZE = jabber.xmlstream.BLOCK_SIZE
 
@@ -1022,6 +1022,9 @@ class JabberClient(jabber.Client):
             self.disconnected(self)
         return received
 
+    def process_loop_hook(self):
+        pass
+
     def process(self, timeout=None):
         log_debug(3, timeout)
         self._incoming_node_queue = []
@@ -1038,6 +1041,7 @@ class JabberClient(jabber.Client):
             else:
                 tm = None
 
+            self.process_loop_hook()
             # tm is the number of seconds we have to wait (or None)
             log_debug(5, "before select(); timeout", tm)
             rfds, wfds, exfds = select.select([fileno], [], [], tm)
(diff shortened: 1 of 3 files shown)


xkollar:
Would you also care to remove unused imports you introduced (like `PIPE` and `STDOUT` in `osad_client.py`)?

PS: Little cleanup in form of fixuping self-cleanup commits like 4a6bbbcbbf68be0b98e48fa23e86696e87e662ce would also be appreciated. (Hint: `git rebase -i` is your friend.)

xkollar:
Any progress?

xkollar:
Ah... never mind... rebased and fixed as b7d1107bbb0ef7ef9e13373405051862b76affb5.

dmacvicar:
I am sorry I could not follow up that fast. I had it on my queue, but I am busy with other stuff.

PR #233, opened 2015-03-04
Avoid a deadlock when changing channel assignments (bsc#918549)

moio:
Avoids a deadlock condition between update_needed_cache and subscribe_server.

In depth information, including reproducer code, is available at:

http://www.moioli.net/progetti/a-postgres-deadlock-bug-case-study/

This post is actually introduced by those two accompanying posts, which may also be relevant for you:

http://www.moioli.net/progetti/deadlocks-in-postgres-101/
http://www.moioli.net/progetti/postgres-deadlocks-debugging-guidelines/

This is the first deadlock bug I try to fix in the Spacewalk codebase, so I am _very_ happy to hear from you about this - is my reasoning correct? Can you reproduce the issue? Is the solution acceptable, can it be improved?

Thanks very much in advance!


sdherr:
Wow, nice write-up. What an unfortunate problem. Yes, I agree that your reasoning is correct and that this is the correct solution. Here's to hoping that adding the explicit lock doesn't open up another deadlock somewhere else.

Committing as b0a27c0f22c2da5054a6543563e97790691c6711.

PR #235, opened 2015-03-05
Add EPEL 7 package key

phurrelmann:


sdherr:
Thanks, committing as d5046d2af5781393407e1ba17613eb0a5e1538e7. Also see in ef4f12d5d401d713822eef83c3870ca19ea4218d, we need to add it to the data directory too for non-upgrades (fresh installations).

phurrelmann:
Thanks, will do next time. Still learning ;)

sdherr:
No worries. Oh, and as for the upgrade script ordering, yes generally new scripts will be <nextnumber>. It only *really* matters if you have order-dependent changes (scripts are executed in numeric order by spacewalk-schema-upgrade) so inserting a new row into rhnPackageKey could probably go anywhere, but if all the new stuff is numbered later it's easier for people running nightly (like me!) to keep track of what we've already applied to our dbs.

PR #238, opened 2015-03-06
Fix spacecmd cli config option nossl

phurrelmann:
Before the fix it triggered the following error:

spacecmd -s spacewalk.example.com --nossl
Traceback (most recent call last):
  File "/usr/bin/spacecmd", line 132, in <module>
    shell.load_config_section('spacecmd')
  File "/usr/lib/python2.7/site-packages/spacecmd/misc.py", line 1001, in load_config_section
    if re.match('^1|y|true$', self.config['nossl'], re.I):
  File "/usr/lib64/python2.7/re.py", line 137, in match
    return _compile(pattern, flags).match(string)
TypeError: expected string or buffer

### DIFF
--- a/spacecmd/src/lib/misc.py
+++ b/spacecmd/src/lib/misc.py
@@ -997,7 +997,7 @@ def load_config_section(self, section):
                 pass
 
     # handle the nossl boolean
-    if self.config.has_key('nossl'):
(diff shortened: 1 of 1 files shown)


tomas-kasparek:
Committed into Spacewalk as bb657fe773fe5497d641915c28bdb2d668347077.
Thank you.

PR #241, opened 2015-03-13
Added a bit more documentation to softwarechannel_setsyncschedule

Djelibeybi:
This at least tells end-users what to search for to find the appropriate schedule format when adding a repo sync schedule.


tomas-kasparek:
Hello Avi, I've rebased your three commits into one as it doesn't make sense to keep this as 3 commits where 2 of them are fixes.
Spacewalk master: 650d7b88cadde5ef170a4f6098f73b45547a8390
Thank you

PR #243, opened 2015-03-20
Improvements for the errata-cache taskomatic job

renner:
This is a patch about improving the errata cache taskomatic job by consolidating the list of tasks returned by `getCandidates()` in a way that only **one** cache update is scheduled per server. This might help to overcome performance problems in large installations where many systems are being managed, i.e. reduce the time needed to process a long queue.

For instance if you update the package list of a server twice in between two runs of the `errata-cache` job, taskomatic will afterwards schedule the cache update twice for the same server. If you register a system, a cache update will even be scheduled 9 times (in my tests) for a single server! When this patch is installed a cache update will only be done once per server for every run of the job.

To see relevant log output you might want to add this to the taskomatic log4j.properties:
    
    log4j.logger.com.redhat.rhn.taskomatic.task.ErrataCacheTask=DEBUG
    log4j.logger.com.redhat.rhn.manager.errata.cache.UpdateErrataCacheCommand=DEBUG

### DIFF
--- a/java/code/src/com/redhat/rhn/common/db/datasource/xml/Task_queries.xml
+++ b/java/code/src/com/redhat/rhn/common/db/datasource/xml/Task_queries.xml
@@ -617,10 +617,10 @@ where snc.errata_id = :errata_id
 </write-mode>
 
 <write-mode name="insert_into_task_queue">
-        <query params="org_id, task_data, cid, earliest">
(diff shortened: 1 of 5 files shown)


tomas-kasparek:
This patch seems to be reasonable, thus I've committed it into Spacewalk.
Thank you for contribution.

PR #245, opened 2015-03-24
tuple assignment should be list in client_config_update.py

rlmoser:
When an = is encountered in a config file value, client_config_update.py
joins the = separated portions together and reassigns to kv, but it makes
this assignment as a tuple rather than as a list, as a result, the line
kv[0] = string.strip(kv[0]) fails due to tuples being immutable.

Fixed by enclosing the assignment in [] to make it a list.


tomas-kasparek:
I've committed your PR into Spacewalk as 02d7a42b0815c0c21c66835ab726c31e5f7b8152.
Thank you for contribution.

PR #246, opened 2015-03-27
Kickstart: install the latest available koan package, not the first one found

moio:
`KickstartScheduleCommand` is supposed to install koan on the host as a prerequisite of deploying a new VM. Current code assumes the package is available in one channel at the latest version, and will install the first one that is cound.

At least in some SUSE channels, that is not always wanted as we have, for example, separate "pool" and "updates" channels with different versions of the same package. This PR compares all available versions of the package and chooses the most recent one for installation.

Note to reviewers: "Refactoring" commits are supposed not to alter the semantics of code and have been separated for clarity.

### DIFF
--- a/java/code/src/com/redhat/rhn/manager/kickstart/KickstartScheduleCommand.java
+++ b/java/code/src/com/redhat/rhn/manager/kickstart/KickstartScheduleCommand.java
@@ -35,6 +35,8 @@ import com.redhat.rhn.domain.kickstart.KickstartSessionState;
 import com.redhat.rhn.domain.kickstart.KickstartVirtualizationType;
 import com.redhat.rhn.domain.kickstart.RegistrationType;
 import com.redhat.rhn.domain.rhnpackage.Package;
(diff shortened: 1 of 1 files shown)


xkollar:
Quite nice work there. What I don't understand though is: why findKickstartPackageToInstall returns null instead of empty map?

moio:
Because the `Map<String, Long>` really represents one single entity, where `String`s are field names and `Long`s are the respective values. It would be possible to create a custom POJO as an additional refactoring, I simply kept the existing structure.

In other words, if the method returned multiple results it would be a `List<Map<String, Long>>` and in that case an empty `List` would make sense in case no result was found. Since it only returns zero or one results, it makes sense to have either one `Map` or `null` IMO.

sdherr:
Looks good, thanks! Committing, but squashing down to one commit (including some line-length-longer-than-92-chars checkstyle fixes). 55e95809a7268f37b9c7678e547a5f4c507bc17c

PR #247, opened 2015-04-10
clone config files without loosing trailing new lines

mcalmer:
Do not loose trailing new lines

### DIFF
--- a/spacecmd/src/lib/utils.py
+++ b/spacecmd/src/lib/utils.py
@@ -764,6 +764,9 @@ def file_needs_b64_enc(self, contents):
     if not contents:  # zero length
         return False
 
+    if contents[-1] == '\n': # trailing new lines
+        return True
+
     text_characters = "".join(map(chr, range(32, 127)) + list("\n\r\t\b"))
     # More than 30% non-text characters -> considered a binary file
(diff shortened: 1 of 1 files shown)


xkollar:
Hi. Could you please provide reproducer for undesired behaviour?

mcalmer:
* create a config channel (e.g. origchannel)
* create a file in the config channel with (multiple) new lines at the end of the file

Example:

-------------------- begin file ------------------------
X=1


--------------------- end file -------------------------
* call spacecmd configchannel_clone origchannel -c clonechannel
* check the cloned file in clonechannel

all new lines got removed. It only contains "X=1" without the new lines at the end.

xkollar:
I see. Thanks for the reproducer.

Playing with it, I have found out that there are more issues:
Not only trailing newline(s), but it seems to me like any trailing white-space is removed.
Also, any leading white-space is removed as well.

Would you try to fix those too please?

sdherr:
Instead of accepting this commit I'm just changing spacecmd to always base64 encode config file contents, because why not, and it fixes all whitespace stripping problems. cd86ea1b8671596f025f378c5c06d6918ce6e7fc

PR #248, opened 2015-04-14
Minor Fixes/Improvements

mmattioli:
Proposal
===

There were a few extraneous whitespace characters as well as use of % rather than .format() when forming strings, these were either removed/replaced and/or updated where I saw them. There was also a missed copyright year update which was corrected.

Reasoning
===

.format() has improved type recognition over % and, in combination with the minor whitespace fixes improves readability. The copyright year update is self-explanatory.

### DIFF
--- a/backend/__init__.py
+++ b/backend/__init__.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2009--2011 Red Hat, Inc.
+# Copyright (c) 2009--2015 Red Hat, Inc.
 #
 # This software is licensed to you under the GNU General Public License,
 # version 2 (GPLv2). There is NO WARRANTY for this software, express or
--- a/backend/__init__.py
+++ b/backend/__init__.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2009--2015 Red Hat, Inc.
+# Copyright (c) 2009--2011 Red Hat, Inc.
 #
 # This software is licensed to you under the GNU General Public License,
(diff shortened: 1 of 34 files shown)


xkollar:
* we update copyright years only when file changed that year besides copyright update...
* In whole project we do not use `str.format`, why star now inconsistently?
* Usage of double newlines as in ` backend/_apache.py` is actually part of pep8... isn't it?

Please update your pull request so it can be considered.

mmattioli:
@xkollar I just pushed two new commits, removed the use of `str.format()` and left the extra spacing. [According to PEP8](https://www.python.org/dev/peps/pep-0008/#id15) extra line breaks should be used "sparingly" and only to separate top-level functions from classes. This PR should have everything I'd like to contribute. Thanks.

xkollar:
Thanks for the update :-). So I looked deeper and there are more sgml files that would benefit from change you did in `backend/satellite_tools/accounts/satwho.sgml`. Do you want to add them into this PR or should I do it?

mmattioli:
@xkollar I just pushed some more commits updating whitespace throughout as many .sgml files I could spot. If there are more I'm missing please let me know and I'll update.

xkollar:
Rebased on master and cleaned a little as 92ac792d6cb7d473a2d9e8223b5aeedf0e717725 and d6b29a0b503d47ca10e0591ddcfe5a66a9f72e63.

Thank you for your contribution `:-)`.

PR #249, opened 2015-04-15
Add package key for Spacewalk 2014 (RPM-GPG-KEY-spacewalk-2014)

phurrelmann:


sdherr:
Thanks phurrelmann, patch accepted. Committing as a96dad600cba9ac03d61cefb537c60b000d17818. While I was there I also added the spacewalk-2015 and Fedora 23 gpg keys. (5e75be88110793eff92fda8e4182530184f1616f)

PR #250, opened 2015-04-24
Update 021-rhnServerNetwork-trigger.sql.oracle

medozas:
IF EXISTS does not exist on oracle. This fails the upgrade from 2.2 to 2.3. Removing IF EXISTS part does work and lets upgrade run smoothly.

### DIFF
--- a/schema/spacewalk/upgrade/spacewalk-schema-2.2-to-spacewalk-schema-2.3/021-rhnServerNetwork-trigger.sql.oracle
+++ b/schema/spacewalk/upgrade/spacewalk-schema-2.2-to-spacewalk-schema-2.3/021-rhnServerNetwork-trigger.sql.oracle
@@ -13,7 +13,7 @@
 -- in this software or its documentation. 
 --
 
-drop trigger if exists rhn_servnet_ipaddr_mon_trig;
+drop trigger rhn_servnet_ipaddr_mon_trig;
 
 create or replace trigger
 rhn_servernetwork_mod_trig


Djelibeybi:
Not sure why there is a DROP TRIGGER line if the SQL does a CREATE OR REPLACE anyway.

medozas:
I think you are missing that names have changed -> rhn_servernetwork_mod_trig (new) != rhn_servnet_ipaddr_mon_trig (old)

Djelibeybi:
I did miss that completely. Sorry for the noise! :)

ggainey:
I need to investigate the best (a?) (any?) way to conditionally-drop in Oracle - otherwise, the schema-update will fail if the trigger is already gone

medozas:
makes sens ... so you have 2 choices then. if the trigger is already gone (which should not be the case, right?): drop an error (ilke I had ;)) or do something that catches...

http://docs.oracle.com/cd/B19306_01/server.102/b14200/statements_9005.htm
https://community.oracle.com/thread/830382

This one suggests doing like (already filled in the right name)

```
declare
  e exception;
  pragma exception_init(e,-4080);
begin
  execute immediate 'drop trigger rhn_servnet_ipaddr_mon_trig';
exception
  when e then
    null;
end;
/
```

medozas:
I referenced an example ...

ggainey:
Change verified - can you submit the change as a PR, so I can make sure you get correct credit for the chg?

ggainey:
Closing, superceded by #251

PR #251, opened 2015-04-24
Update 021-rhnServerNetwork-trigger.sql.oracle

medozas:
This permanently fixes https://bugzilla.redhat.com/show_bug.cgi?id=1215151 in a proper way. 1. it does not "IF EXIST" the trigger, since that does not exist in oracle. 2. it does indeed check if the trigger exists, and if so, removes it, otherwise just continues, as the trigger does not exist anymore.


ggainey:
Accepted w/commit 0955ef5a2e487d5c75601fedcca61d946f0508af - thanks for the catch and the fix!  Added you to the contributors-list for spacewalk-2.3 :)

PR #252, opened 2015-05-04
set primary netinterface correctly during registration and on hardware refresh

mantel:


moio:
LGTM, but I have limited Python experience especially in styling.

For example, is it OK to name a variable `primif` or would something like `primary_interface` be preferable? Maybe it's just my Java background :smile_cat:

mantel:
Well, I do not have any Python experience, but I thought if variable names like "h" are ok, longer ones should be ok as well :)

sdherr:
I am opposed to the idea of automatically overriding a user-set primary interface any time a hardware refresh happens to come in over a different interface. If the user has set a primary interface (through the system details "Hardware" tab) we should trust that they actually know what they're doing.

What I'll do instead is 1) accept this commit and 2) add another one that modifies the queries so that we only ever set a primary interface if one is not already set for this server.

It turns out the only time that the "primary interface" actually matters is when displaying things in the webui. So for example if you are re-provisioning a machine and on the advanced settings page you change the Network Connection setting from "first available interface" (default) to a specific interface, then the selected-by-default interface is the "primary interface". So it's more of a suggestion than anything. Or another example is that on the Hardware page for a server there's an "IP Address" field, which is really the IP Address of the "primary interface". So I think it makes sense and is safe to set the interface of the incoming connection as "primary" if none is set, and then never change it. That way if the user decides they really want a different interface to be primary (maybe they care about what the IP Address for that server is on a different subnet than the one that Spacewalk is on?) we won't overwrite their settings.

sdherr:
Oh, never mind, it looks like you have already set it up so that the primary interface will not get overwritten during a hardware refresh. I misread the code the first time. Never mind, my objection is nullified.

Committing as 62f41d6956da75207185d0a7b26769710fa39caf and 7c2c512d7376b417acc09fd3a3ff1c0ff53b18f1. Thanks!

mantel:
I would like to note that displaying things in the web UI is NOT the only situation where the primary interface happens: If you have clients that use ssh-push or ssh-push via tunnel the primary network interface is very important! If the value just gets overwritten by a hardware refresh, some machines all of a sudden are no longer reachable after such an action! So it is not just cosmetic.

PR #253, opened 2015-05-12
Update README.md

mmattioli:
Proposal
===
* better use of markdown headers and links
* removed unnecessary verbiage from resources section
* capitalized resource names and removed stray comms from unordered list

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,9 @@
-Spacewalk [![Build Status](https://travis-ci.org/spacewalkproject/spacewalk.svg?branch=master)](https://travis-ci.org/spacewalkproject/spacewalk/builds)
-=========
+# Spacewalk [![Build Status](https://travis-ci.org/spacewalkproject/spacewalk.svg?branch=master)](https://travis-ci.org/spacewalkproject/spacewalk/builds)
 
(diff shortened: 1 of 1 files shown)


xkollar:
Can you explain how this is better use of markdown? I personally do not like hash-style titles for levels 1 and 2. Also I do not see point in expanding links. I am against accepting this pull-request should it contain those two changes.

mmattioli:
Using `= `or `-` for titles only handles level 1/2, hash-style titles work for deeper levels, why would you want to mix? I can understand not wanting to expand the links but I felt the `README.md` is short enough that it doesn't really matter that much, it makes it easier to see the link right there inline rather than have to look elsewhere to make sure the link is correct not to mention you have to come up with a variable name every time you want to move the links elsewhere so that they're referenced properly. If you want me to change the links back I can do that but you really don't think hash-style titles is more proper than using `=` or `-`  given that the latter only supports level1/2 and the former supports more levels?

xkollar:
I am afraid that those arguments left me unconvinced. You are just trying to impose your taste onto single file. I am not telling it is wrong, just that I do not see reason to accept this pull-request in its current form. Sorry about that.

PR #254, opened 2015-05-12
Change Activation Key Child Channels from multiple select to checkboxes

hustodemon:
In the dialog "Update Activation Key" -> tab "Child Channels" the
multiple select box was not really intuitive to use (need for holding
CTRL for multiple selection). This patch replaces it by checkboxes for
child channels as shown on attached screenshots.

after1
after2


sdherr:
Looks and works great, thanks! Committing as 4b2ee50aaf85484d9bbf857e64326b56c92b85d7.

PR #255, opened 2015-05-20
dnf is default on fedora22, so require it instead of yum-rhn-plugin

Pajinek:
bad dependency for fedora 22


xkollar:
Rebased on master as ecfcbd72912f9916822da5179654d339ace05185. Thank you for your contribution.

PR #256, opened 2015-05-20
Modify Oracle channel names to fit into the 64 character limit.

Djelibeybi:
Reported via the Spacewalk mailing list.


tlestach:
Thank you for the PR.
I doubled the size of rhnContentSource.label in facb6866380eb36e7791340614f765b004e564bd, so there's no need to shrink the names any more. I'm rejecting the PR.

PR #257, opened 2015-05-25
Drop code used by the Perl stack to 'trickle' OSAD

moio:
This was a hack to begin with, Perl side was removed, this patch removes the Python side as well.

See the following code to see where it was used (I could not find any other references):

https://github.com/spacewalkproject/spacewalk/blob/SPACEWALK-2.1/web/modules/rhn/RHN/DB/Scheduler.pm#L40


jdobes:
Makes sense to me. Rebasing to master.

PR #258, opened 2015-05-25
ErrataManager: fix stack update case

moio:
In the case of multiple systems with different stack update erratas, previous code would apply all erratas to all systems which is not correct.

This lead to the failure of the following testcase:

com.redhat.rhn.manager.errata.test.ErrataManagerTest.testPackageManagerErratas()

This patch treats stack updates as normal errata (apart from the ordering), which makes the test pass.

AFAIU this is a regression from the commit:

https://github.com/spacewalkproject/spacewalk/commit/eec22d4bd2db70a1dd1f569cf53fdbb99724f854


jdobes:
Makes sense, fixes the problem and makes code more readable. One more thing, please review the line I marked.

moio:
Comment addressed, thanks for reviewing!

jdobes:
Thanks, I merged it into master as one commit: 1fa28b41f81bed9f20987f6622a884008231fcc7

PR #259, opened 2015-05-26
Get rid of IE7 compatibility mode enforcement

hustodemon:
This was needed because of 'editarea' component which is no longer used (replaced by ace editor in [here]). The IE7 compatibility mode is however incompatible with [bootstrap] which results in broken layout on pages which enforce it (see pictures).

### Screenshot of an affected screen before:
with_ie7_compat
### and after:
wo_ie7_compat

[here]: https://github.com/spacewalkproject/spacewalk/commit/7373be07f67933f784e322a2b168dbabdaf0f967
[bootstrap]: http://getbootstrap.com/getting-started/#support-ie-compatibility-modes


jdobes:
Reproduced on IE 11, your patch helped. Merged into master as b647ceb89ec16487c5c21dc01dd6057f1ebdf3ef. Thank you.

PR #260, opened 2015-05-26
Unify profile creation/update with one submit button instead of two.

lucidd:
This removes the `Upload File` button on the `Create/Update Kickstart Profile` page which does the same thing as the `Create` or `Update` button but uses the file input instead of the textbox. With this patch the `Create` and `Update` button will use either the file input or the textbox depending on which one is provided. If both are provided it will show an error telling the user to only provide one of the two.

#### Before
before_ff

#### After
after_ff

### DIFF
--- a/branding/css/spacewalk-fixes.less
+++ b/branding/css/spacewalk-fixes.less
@@ -12,3 +12,7 @@
         margin-bottom: 0;
         padding-top: 7px;
 }
+
+input[type="file"].form-control {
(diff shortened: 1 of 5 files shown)


xkollar:
Rebased as 9c78ecdd14255309516c850b8edb67a579651a26, 9aad0a7c9c34ed66bccb88fb17a76c1538921bd8. Thank you.

PR #263, opened 2015-06-01
Bz1226986

Pajinek:
accept options from plugin configure file spacewalk.conf

# cat /etc/dnf/plugins/spacewalk.conf 
[main]
enabled = 1
gpgcheck = 1
timeout = 150

[fedora-22-x86_64]
timeout = 50
gpgcheck = 0


MichaelMraka:
Merged. Thanks.

PR #264, opened 2015-06-03
Add a spacewalk-branding-devel package to install LESS files for development

moio:
Sometimes we just need the LESS source file around for hacking and testing.

This PR creates a small package to distribute it.


tomas-kasparek:
Looks reasonable, committed into Spacewalk master. Thank you

PR #265, opened 2015-06-03
OSAD: log error stanzas

moio:
This PR has two refactoring commits, which are only about removing cruft, and two commits to enable better logging in case of errors.

We had a customer reporting random failures in applying actions to a large number of OSAD clients. Turned out that OSAD requires all clients to have different credentials in /etc/sysconfig/rhn/osad-auth.conf, in fact as soon as two clients have the same file they will conflict but unfortunately, that would not be reported clearly by the logs.

By XMPP specification, in fact, server sends clients a special "conflict" stanza and closes the connection of one of the clients with the same JID (http://xmpp.org/rfcs/rfc3921.html#session):

"If there is already an active resource of the same name, the server MUST either (1) terminate the active resource and allow the newly-requested session, or (2) disallow the newly-requested session and maintain the active resource. Which of these the server does is up to the implementation, although it is RECOMMENDED to implement case #1. In case #1, the server SHOULD send a <conflict/> stream error to the active resource, terminate the XML stream and underlying TCP connection for the active resource"

(Jabberd actually implements case #1)

Solution was simply to regenerate /etc/sysconfig/rhn/osad-auth.conf by deleting it, I just want to make sure such problems are more clearly visible next time!

### DIFF
--- a/client/tools/osad/src/jabber_lib.py
+++ b/client/tools/osad/src/jabber_lib.py
@@ -979,7 +979,17 @@ class JabberClient(jabber.Client, object):
                 'subscription'  : subscription,
                 'ask'           : 'subscribe',
             }
-            self._subscribe_to_presence(jid)
+
+            # subscribe this node to the jid's presence
+            log_debug(4, jid)
+            stripped_jid = self._strip_resource(jid)
+            presence_node = JabberPresenceNode(to=stripped_jid, type="subscribe")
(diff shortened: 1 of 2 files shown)


tomas-kasparek:
I've committed your changes into Spacewalk master as they make perfect sense. Thank you for contribution!

PR #266, opened 2015-06-09
Fix Debian packages file

rpasche:
Workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1226329
If the version is preceeded with operators, add an extra space between operator and version to be parseable.

I have to tell, that I was not able to fully test this (in spacewalk), but made a small test script. It should work, to change a version line (e.g. "<<1.3.4-5" to "<< 1.3.4-5")

### DIFF
--- a/java/code/src/com/redhat/rhn/taskomatic/task/repomd/DebPackageWriter.java
+++ b/java/code/src/com/redhat/rhn/taskomatic/task/repomd/DebPackageWriter.java
@@ -43,7 +43,7 @@ public class DebPackageWriter {
     private static Logger log = Logger.getLogger(DebPackageWriter.class);
     private String filenamePackages = "";
     private String channelLabel = "";
-    private Pattern versionRegex = Pattern.compile("^([<>=]+)\s*(.*)$");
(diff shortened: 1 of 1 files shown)


tlestach:
Note for whoever will review this PR: As stated - this a workaround, not a proper fix. See the Bug 1226329 for more information.

rpasche:
It's true, that this is really only a workaround, but it would help a lot to get around the error I have with Debian 8. Now, I have to re-create the Packages.gz after the taskomatic daemon has created the original Packages file. OK...not really a big deal, but annoying. Also I cannot imagine, how long it will take to fix the root cause of this problem within spacewalk.

Thubo:
Any news on this?

rpasche:
Not as far as I know...

ggainey:
rpasche, Thubo - Could you look at PR#399, and see if it is a 'correct' fix for this problem?

rpasche:
Hi @ggainey ,

just commented the pull request...don't know, when I have time to test it.

ggainey:
@rpasche @Thubo - I'm going to close this PR in favor of the work @tlestach is doing in #399 - would rather have this fixed correctly than hack around the issue, and looks like we may be close to that.

PR #268, opened 2015-06-11
When creating a channel, use x86_64 as the default architecture

dmacvicar:
Most users will have to click on IA-32 just to change it to x86_64.

For the future: At some point it would be useful to evaluate if checksum should be set to something higher than sha1. As the page automatically changes the checksum to the parent channel, setting the default would not affect older distros.


jdobes:
This change sounds reasonable. But your patch does not make any visible change. I found out it's caused by line:

https://github.com/spacewalkproject/spacewalk/blob/master/java/code/webapp/WEB-INF/pages/channel/manage/edit.jsp#L153

It overrides default value for some reason. Please check how to safely change/remove this line and update PR.

dmacvicar:
I am sorry, I messed up something here. I will do a new single PR.

PR #271, opened 2015-06-12
New Channel: Fix setting the default arch/checksum when selecting back Parent: None

dmacvicar:
* If the checksum is set to null because of setting back Parent to none, no metadata will generated
  and no feedback is given to the user. This prevents this problem.
* The default still comes from the Action default values, so no duplicates of default values are
  kept in the view.

While at it:

* Refactored the javascript generation to use standard js objects instead of a custom
  encoding using ':' and ',' that required unnecessary joining and splitting.
* Use the jquery functions to remove onChange tag and keep the js unobtrusive. Step
  needed to get rid of all inline Javascript in the future.

Demo of the fix: (demo taken in our maintenance branch, based on 2.1, that has not the architecture compatibility map, and therefore is a simplified version of this patch)

5685ef3a-1017-11e5-8da9-856587d4483d

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/action/channel/manage/EditChannelAction.java
+++ b/java/code/src/com/redhat/rhn/frontend/action/channel/manage/EditChannelAction.java
@@ -54,6 +54,8 @@ import org.apache.struts.action.ActionMapping;
 import org.apache.struts.action.ActionMessage;
 import org.apache.struts.action.ActionMessages;
 import org.apache.struts.action.DynaActionForm;
+import org.stringtree.json.JSONWriter;
+
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


moio:
I am OK with this PR in the sense that it fixes the original problem and it is an improvement over previous code, although I must say I dislike:

 1. having `<c:forEach>` tags in Javascript code;
 2. constructing HTML in Javascript.

For 1. I would suggest to either inline data in HTML using `data-` tags as strings/numbers, if it makes sense, or as JSON, and have Javascript parse that.

@dmacvicar, I guess the above is not part of this PR because it's out-of-scope, correct?

Point 2. might be more of a personal preference, but I prefer having all HTML structure in the JSP file. In this case I would probably have all `<option>` tags already present in the JSP page and hidden as needed, unhiding them from Javascript code depending on user actions. Comments welcome.

dmacvicar:
* <c:forEach> was already used in the original code. It avoids to fill the JS array with data that would not be used. One could as well do it in the action and just pass a more complex JSON object.

Right now, we iterate the map in JSP and add a JSON object on each iteration. One could as well just pass the whole map from the beginning as JSON.

* I think the idea of using data-attributes is a good idea. I could give it a try.

dmacvicar:
Ok, looks better. I suggest keeping the forEach as it is part of the original code and I prefer to tackle it as a separate issue.

moio:
:+1:

dmacvicar:
Superseded by https://github.com/spacewalkproject/spacewalk/pull/276

PR #272, opened 2015-06-13
auto-healing for duplicate jabber ids

mcalmer:
In case 2 systems have the same jabber id, one will get a conflict message and will be dropped from the server. (E.g. cloned image)
Raising NeedRestart force osad to recreate osad-auth.conf with new credentials and authenticate again at the server.


tomas-kasparek:
Comitted into Spacewalk. Thank you.

PR #276, opened 2015-06-26
Improvements to the new Channel form

dmacvicar:
This PR adresses the following things (in separate commits)

* Some refactoring on the Javascript to be unobtrusive
* Removes the checksum type None as discussed in https://www.redhat.com/archives/spacewalk-devel/2015-June/msg00008.html
* When no parent channel is selected, a default architecture is selected (which is defined on the action side, not in the view)
* The default architecture is x86_64
* Supersedes: https://github.com/spacewalkproject/spacewalk/pull/268 and https://github.com/spacewalkproject/spacewalk/pull/271

# Before

before

# After

after-1

out

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/action/channel/manage/EditChannelAction.java
+++ b/java/code/src/com/redhat/rhn/frontend/action/channel/manage/EditChannelAction.java
@@ -98,6 +98,11 @@ public class EditChannelAction extends RhnAction implements Listable<OrgTrust> {
     public static final String CHANNEL_ARCH = "channel_arch";
     public static final String CHANNEL_ARCH_LABEL = "channel_arch_label";
 
+    public static final String DEFAULT_ARCH = "channel-x86_64";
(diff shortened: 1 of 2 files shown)


jdobes:
There is problem with the default arch commit. It causes bug in channel cloning. Selected architecture for new cloned channel is not same as architecture of original channel.

dmacvicar:
Before the arch was passed as channel_arch_label. I setted the form variable instead and that broke cloning and probably editing.

So in commit 8668939970a000b2b608fb1569e77ea762d6e167 I reverted now that change, and now I set the channel_arch_label in the "New channel" case, producing the same effect without changing the original code paths.

As I could not reproduce the cloning problem here, I hope this is the "way". Feedback & guidance appreciated.

jdobes:
It fixed the problem. I added your changes into master. I just squashed last two commits and changed "ctx.getRequest()" to "request".

We removed None checksum support only by removing select option. Are there any other places in code handling None checksums? If yes, we should clean them as well.

Thank you.

PR #277, opened 2015-06-30
use unverified context method for ssl self-signed connection

Pajinek:
the fix of following bug https://bugzilla.redhat.com/show_bug.cgi?id=1237219


jdobes:
We fixed it different way.

PR #285, opened 2015-07-22
Use reposync logs to show status of last reposync in WebUI

jdobes:
## Enhancements:
* Parse last reposync log and write information into progress bar on the page.
* Detect running reposync for given channel and notify user.
* Add link to reposync logs (same as on Channel Details page).

## Before:
before

## After:
after1
after2


jdobes:
Merged.

PR #286, opened 2015-07-23
Remove test namespace

xkollar:
Discovered by calling "api.getApiNamespaces".
Contains all the "handy" function one might want, some of which does not even work...

* addition
* envIsSatellite
* hashChecking
* multiplication
* singleIdentityFunction


jdobes:
Merged into master, thanks.

PR #287, opened 2015-07-28
osad: re-send subscription stanzas after a while

moio:
The [XMPP protocol](http://xmpp.org/rfcs/rfc3921.html) defines [a state machine for the subscription](http://xmpp.org/rfcs/rfc3921.html#substates) of a client to a contact's state - this is basically a series of intermediate states clients have to go through before getting any notification of another client's availability. We use this to subscribe RHN clients to `osa-dispatcher`'s messages, so clients are expected to go through all states up to `Both` at registration time or soon afterwards.

If for any reason presence stanzas are lost in the process (due to network failures, bugs in osad code, bugs in jabberd etc.) RHN clients might be stuck in one of two limbo states waiting for server to acknowledge a message that it has never received.

The only solution is for clients to retry sending those messages, as `osa-dispatcher` might not even know about client's addresses at this stage - the XMPP standard explicitly states that they should be hidden from its roster:

> 8.2.  User Subscribes to Contact
> [...]
> adding a roster item for the contact to the user's roster, with a state of "None + Pending In" as defined under Subscription StatesSubscription States, however a server SHOULD NOT push or deliver roster items in that state to the contact

Thus this patch adds retrying of subscription in case the state is one of the two limbo states and more than one minute elapsed.

### DIFF
--- a/client/tools/osad/src/osad.py
+++ b/client/tools/osad/src/osad.py
@@ -247,6 +247,11 @@ class Runner(jabber_lib.Runner):
             config = self.read_config()
             raise jabber_lib.NeedRestart
 
+        # make sure that dispatchers are not stuck in state [none + ask] or [from + ask]
+        # for too long. This can happen, for example, if a "subscribe" presence stanza
+        # gets lost - in that case re-send it
+        client.unstick_contacts(self._dispatchers)
+
         # if rhn_check is running or the last one failed, check more often
         if (client._rhn_check_process is None) and (client._rhn_check_fail_count < 1):
             client.process(timeout=180)
--- a/client/tools/osad/src/osad_client.py
+++ b/client/tools/osad/src/osad_client.py
@@ -36,6 +36,7 @@ class Client(jabber_lib.JabberClient):
         self._config = {}
         self._rhn_check_process = None
         self._rhn_check_fail_count = 0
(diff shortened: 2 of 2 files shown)


moio:
Please also note that deleting the jabberd database will trigger the subscription process from the beginning in all clients. We have reports from customers that servers with a sizable amount of clients (hundreds) this typically ends in some of them ending in a limbo state.

@sdherr: there's a [comment from you](https://github.com/spacewalkproject/spacewalk/commit/eadd67e375d25363d0d21fb22656b01b96716e3d#diff-82b20e17dda0a97d2aea7ebe36c0ba7bR152) in osa-dispatcher that says that we "blow away" the jabberd database but I could not find out where we do that and why. In SUSE Manager we do not do that, so I would like to understand the deeper reason (maybe it's addressed by [my PR in upstream jabberd](https://github.com/jabberd2/jabberd2/pull/99)?).

tomas-kasparek:
Looks reasonable. Merged into Spacewalk.

vinzent:
@moio RedHat unofrtunatly has included a delete of the jabberd database in the satellite init scripts (at least up to 5.7)

moio:
@vinzent oh, thanks for finding out! :smile:

PR #291, opened 2015-08-05
Modified setup string to make it clearer for Oracle RDBMS

Djelibeybi:
Using the Oracle Instant Client implies that most connections will use the ezconnect syntax, i.e. \\host[:port]\database_name. In this syntax, we require the Global Database Name (db_name.db_domain) and not the legacy SID value.

For local XE installs, the global database name and SID are the same value, i.e. "XE", so //localhost/XE works without a tnsnames.ora.


tomas-kasparek:
Makes perfect sense. Committed into Spacewalk.
Thank you

PR #295, opened 2015-08-22
Show a descriptive message on reboot

vinzent:
Right now only "The system is going down for reboot in 3 minutes!" is shown.
Users on the shell only can guess why the system reboots.

This patch adds a descriptive message to the shutdown command. It will
print additionally 'Reboot of system "$hostname" initiated by
spacewalk reboot action'

### DIFF
--- a/client/rhel/rhn-client-tools/src/actions/reboot.py
+++ b/client/rhel/rhn-client-tools/src/actions/reboot.py
@@ -31,12 +31,13 @@ def reboot(test=None, cache_only=None):
 
     pid = os.fork()
     data = {'version': '0'}
+    reboot_message = 'Reboot of system "' + os.uname()[1] + '" initiated by Spacewalk reboot action.'
     if not pid:
         try:
             if test:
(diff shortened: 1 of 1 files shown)


jdobes:
Seems as simple nice-to-have enhancement. Just few notes:
* Isn't printing hostname a bit redundant information here? Also it's already mentioned in "Broadcast message from root@$hostname (Fri 2015-08-28 10:50:02 EDT):" string.
* "spacewalk" should be with big 'S' letter to be consistent with other strings. There should be dot at the end of the sentence, all other sentences of reboot announcement have some interpunction.

vinzent:
i have adressed the spelling with the last update.

I review the hostname in the message on monday. IMHO there was some reason to include it. :)

vinzent:
@jdobes EL5 does not print the hostname. At least for EL5 systems it's not redundant.

jdobes:
All right then. Merged as b518880c6ee924baa32acca1ac53f00f43c5d063 . Thank you!

PR #296, opened 2015-08-29
Make script reusable for more projects

xkollar:
Hi guys.

Long time no see `:-)`... Aaaanyway... This PR si about making `github-fetch-pullrequest` more versatile. While the change might not be of particular value for Spacewalk itself, it might be interesting to separate the script into its own repo. (That way it would be simpler to use for other projects... like transitions and maybe even some other people might like to use it.) So please consider creating separate repo under Spacewalk organization and moving script in question there.

Good luck and maybe see you some time.

### DIFF
--- a/scripts/github-fetch-pullrequest
+++ b/scripts/github-fetch-pullrequest
@@ -11,12 +11,13 @@ Usage:
 
 import git
 import os
+import re
 import simplejson
 import sys
 import urllib
 
-USER = "spacewalkproject"
-REPO = "spacewalk"
+USER = None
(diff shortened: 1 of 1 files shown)


tlestach:
Accepted as 8a6613a63945ea2b197bf95c254c338135d32478. Thank you for contribution!
I agree this script is not Spacewalk specific. It would deserve a separate git repo, however, I'm not sure, whether spacewalkproject organization is the right place for such repo.
What about packaging the script and getting it into fedora?

PR #298, opened 2015-09-08
Fix pagination for `<rhn:list .../>`

hustodemon:
Change https://github.com/spacewalkproject/spacewalk/commit/65ed77d32f438c516e8779d5fdb6611756a2d333#diff-a53224fadf9a06ea207bc7d70fa8b117R136 disabled pagination for lists created using `<rhn:list .../>` tag.

The bug was caused by changing the string `"First"` to `"First Page"` in `ListDisplayTag`. This string served as a pagination-controlling parameter in the request. However, the request handling logic (in `RequestContext`) still relied on the string value `"First"`.

Fixed by unifying the strings and putting them into a single enum.


jdobes:
Merged into master. Thank you.

It would be nice to have these strings for all types of lists on one place and put them out of the code one day.

PR #301, opened 2015-09-25
Clean up base classes for tests

renner:
This patch cleans up `RhnJmockBaseTestCase` so that the `setUp()` method would actually do the same thing as the `setUp()` in `RhnBaseTestCase`. Unreferenced methods are further removed and the contents of `setUp()` is moved into `TestCaseHelper` for de-duplication.


ggainey:
Accepted as 5b18c57d 7ecf7814 and 84d8996f - thanks for the patch!

PR #302, opened 2015-09-25
Hibernate java.lang.ClassCastException: java.lang.Boolean cannot be cast to java.lang.String

dmacvicar:
rename ChannelProduct#beta to ChannelProduct#betaMarker

Following bean naming conventions or spending too much time choosing the right name may sound "picky", but these things always backfire.

ChannelProduct#beta property makes you think it is "is the product beta?", but it is a string that suggests the beta status (it can be 'Y'/'N' but what if it is 'T'?). That field should have been named betaMarker, betaFlag, whatever.

To make life easier a boolean helper was created isBeta()¹ and a boolean setter, which because of Hibernate complaining was named "setMyBeta()"²

Problem is:

* ¹ isBeta clashes with the other property setBeta/getBeta

In Java 7 Class.getDeclaredMethods() does not guarantee order anymore. (http://bugs.java.com/bugdatabase/view_bug.do?bug_id=7023180)

So running tests with OpenJDK, Hibernate will get wrong method for the property
"beta" and the following error will happen:

    java.lang.ClassCastException: java.lang.Boolean cannot be cast to java.lang.String

* Even if the badly named setMyBeta() was added, it was not used!. Right. All code goes over setBeta, passing again the 'Y'/'N' value, which is error prone.

So this PR proposes the following:

* Rename the badly named method setBeta/getBeta to setBetaMarker/getBetaMarker
* Rename isBeta and setMyBeta to isBeta/setBeta(bool) to avoid the clash in ¹ but also add some consistency
* Actually USE setBeta in the boolean form (it was not)
* Make the string versions (betaMarker) private.

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/channel/ChannelProduct.hbm.xml
+++ b/java/code/src/com/redhat/rhn/domain/channel/ChannelProduct.hbm.xml
@@ -14,7 +14,7 @@ PUBLIC "-//Hibernate/Hibernate Mapping DTD 3.0//EN"
 
         <property name="product" column="PRODUCT" not-null="true" type="string" length="256" />
         <property name="version" column="VERSION" not-null="true" type="string" length="64" />
-        <property name="beta" column="BETA" not-null="true" type="string" length="1" />
+        <property name="betaMarker" column="BETA" not-null="true" type="string" length="1" />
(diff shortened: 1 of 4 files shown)


tlestach:
This makes sense. Accepting as 58c63bfdd5fffd4331895d23c6ced65e542f9a2d and 38618ef53ca05de2c47add19acab1c750c1a853c. Thank you, guys!

PR #303, opened 2015-09-26
fix schema migration error - trigger must be created after the table

mcalmer:
Take care that triggers are created after the table.


jdobes:
Merged into master, thanks!

PR #305, opened 2015-10-01
Pwayper close unused check connection statement

PaulWay:
This is a slightly more verbose patch for the problem described in Red Hat bug 1266216:

https://bugzilla.redhat.com/show_bug.cgi?id=1266216

Basically, the problem seems to be that in some circumstances, the "select 1" statement can cause the Postgres server to keep a statement cursor open.  This ties up extra resources in the database, which can lead to connection problems, memory problems and more.


PaulWay:
Could also just be junk = c.fetchall() to do less work.

tlestach:
According to the referred BZ, this patch did not help. The real fix for the BZ is already in upstream.
Closing.

PR #308, opened 2015-10-06
fix rhnChannel instance has no attribute 'get'

twouters:
getChannels.channels() returns a list of rhnChannel objects, which don't have a 'get' method.
This fixes an AttributeError exception while running `spacewalk-channel -b`.


tlestach:
Correct.
Accepting as d7a91117e7ee5e2ae35d35db51d6a8c63a29a3d7. Thank you!

PR #309, opened 2015-10-22
Fix pwstrength-bootstrap license

moio:
From https://github.com/ablanco/jquery.pwstrength.bootstrap

"Dual licensed under the MIT and GPL licenses. You can choose the one that suits your purposes better."


jdobes:
Merged as b62a2dc5e56203c2c01ad4f6114e5aae8ede7ab5 . Thanks.

PR #310, opened 2015-10-26
Fix equals & hashCode in VirtualInstance and consequences

hustodemon:
- Get rid of using ID in equals & hashCode in VirtualInstance.
- Remove incorrect all-delete-orphans cascade from server->guests
  mapping. In some cases, this cascade can remove VirtualInstances that
  still point to a registered guest system. Programmatic deletion of the
  orphans in appropriate case was implemented instead.
- Adjust tests.


hustodemon:
This change actually revealed some other interesting bugs which need to be handled. Closing this PR for now.

PR #311, opened 2015-11-04
Add option to softwarechannel_setorgaccess for protected sharing of channels

photoninger:
Working with multiple organisations which do not share all channels with each other it is convenient to extend the softwarechannel_[s|g]etorgaccess commands:
I added the option --protected to enable channel sharing only for the given organisations to software_channelsetorgaccess.
software_channelgetorgaccess now reports for protected channels for all trusted organisations if they are enabled or disabled.
And I added software_channel[s|g]etorgaccesstree to work with whole channel trees. Most of the time you want to have the same sharing options on the whole channel tree of a base channel.

I could not figure out why rel-eng/packages/.readme was changed...

### DIFF
--- a/rel-eng/packages/.readme
+++ b/rel-eng/packages/.readme
@@ -1,3 +1,3 @@
-the rel-eng/packages directory contains metadata files
-named after their packages. Each file has the latest tagged
-version and the project's relative directory.
+Files in this directory are updated using tito tag invoked from
+directories of individual packages.
+Do not modify these files manually without knowing what your've doing.
--- a/spacecmd/src/lib/softwarechannel.py
+++ b/spacecmd/src/lib/softwarechannel.py
@@ -1593,16 +1593,13 @@ def complete_softwarechannel_setorgaccesstree(self, text, line, beg, end):
     return tab_completer(self.list_base_channels(), text)
 
 def do_softwarechannel_setorgaccesstree(self, args):
-    if not len(args):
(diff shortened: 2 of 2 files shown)


jdobes:
Hello, I marked 2 lines, build failed because of them. Please look at them.

What's the point of "I dont know what i am doing." commit? :-)

photoninger:
@jdobes in line 1605 I used do_softwarechannel_setorgaccess as template, there we have also two times "if not len(args)": 
The first time args contains all arguments, the second time args only contains the remaining args (channel names) after extracting the options. 
So if you call do_softwarechannel_setorgaccess without any arguments the first one will be used, but if you call do_softwarechannel_setorgaccess only with options but without any channel names the second one will be used. 
Perhaps the first one can be deleted.

photoninger:
@jdobes the "I dont know what i am doing." commit was because git did not allow me to do "git rebase upstream/master" because of unstaged changes in rel-eng/packages/.readme. I did not touch this file and a "git checkout -- rel-eng/packages/.readme" did also not revert the changes.

tlestach:
@jdobes, if you'd accept the PR, please add the bug number 1277994 to the commit messages. Thank you.

jdobes:
It does what it says, merging as one commit - 95dd6678b33ebf81bb44ad6b6d5533e642f4b8b6 . Thanks.

PR #312, opened 2015-11-06
BugFix: skip similar tasks only if task is 'single threaded'

ncounter:
If a task is defined as `single threaded`, any other calls of this task are skipped if one of that type is already running. While checking if a task is single threaded, `TaskoJob.java` does an incorrect check returning always true and skipping all task but one for each type of task. This should not happen for `reposync` task, because it is not a single threaded task, and all `reposync` tasks should be run even if any other is running.

The solution: revert the behaviour of `TaskoJob.java` to the original one and refactor it.

The bug was introduced by removing the casting in the commit https://github.com/spacewalkproject/spacewalk/commit/f3d9e83a1d28b917c20aa6fcd6d92d55ee8d91cc

### DIFF
--- a/java/code/src/com/redhat/rhn/taskomatic/TaskoJob.java
+++ b/java/code/src/com/redhat/rhn/taskomatic/TaskoJob.java
@@ -17,6 +17,7 @@ package com.redhat.rhn.taskomatic;
 import com.redhat.rhn.common.conf.Config;
 import com.redhat.rhn.common.hibernate.HibernateFactory;
 import com.redhat.rhn.taskomatic.task.RhnJob;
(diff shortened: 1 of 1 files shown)


jdobes:
You are right, the current behaviour is incorrect.

Also reported as a bug https://bugzilla.redhat.com/show_bug.cgi?id=1278988

jdobes:
Since you are author of the original change, could you take a look on this, @tlestach ? Thanks.

tlestach:
Ups. I must have missed automatic optimization corrections of my IDE. :-(
Anyway, thank you, ncounter! Good catch!
Accepting as ac450714588297a5198af3e5d819fd5004e3813b

PR #314, opened 2015-11-17
BugFix: remove inconsistency action description

ncounter:
There is an inconsistent action description on `package install` and `package upgrade` since the action at the core of the functionality is the same, but the action description says `install` and the package tab says `update`.
This pull request remove this inconsistency and join the separate tab creation on `package remove` to a unique generic tab because the title already explains the action and the tab contains just a list of packages.

BEFORE:
screenshot from 2015-06-19 09 45 28

AFTER:
screenshot from 2015-11-13 07 36 10


jdobes:
Useful enhancement, thank you. Took me a while to realize the label "Package Install/Upgrade" will show only for newly created actions, not for already existing. :)

bca3f6122846687ed1d1ae388a6712e71de08a80

PR #316, opened 2015-11-24
Fix: sort channel list by name

ncounter:
This change extracts channels from the database ordered by name.


ggainey:
Accepted as 0094dcae - thanks for the patch!

PR #317, opened 2015-11-25
When deleting a server, delete all associated rhnSet entries

moio:
Previously only entries from the same Org were deleted, but this left dangling references that were not cleaned up.


ggainey:
Committed as 550be0ed - thanks for the patch!

PR #318, opened 2015-11-25
Fix: prevent return null on merging path slices

ncounter:
This changes the merger method for path pieces to prevent returning `null` (i.e. when path contains `///`).
The javadoc changed because it was a little misleading about the <i>url path</i> case that never happen in that method.


ggainey:
Accepted as 2f19c70e thanks for the patch!

Also, I added a junit test for this method - see ead3277c

PR #319, opened 2015-11-25
Unify system->proxy tabs

mateiw:
Merge the two tabs under system->proxy into a single one:
* changed system->proxy to point to page proxyclients 
* added proxy version to the proxyclients page
* removed unused proxy page

proxy_tab_change_one_client
proxy_tab_change_no_clients


jdobes:
I agree, it's better now with single page. And having page called "Activation" which says only the version and cannot activate anything is confusing.

I merged it to master, added commit deleting unused links instead of just commenting them. I also cleaned other than English resource changes in "update strings" commit as we are trying different workflow currently - to avoid problems when syncing languages in git with Spacewalk Zanata instance.

PR #321, opened 2015-11-25
Refactor the Satellite entitlement

moio:
The Satellite entitlement is not a real entitlement and has no use we could imagine, so we propose to simply remove it.


ggainey:
Accepted as fa1b0074 3da96f7f and 709115bb thanks for the patch!

PR #322, opened 2015-11-25
Virtual system list fixes

moio:
While working with Virtual Systems we found out some discrepancies that are addressed by individual commits in this PR.

Comments welcome!

### DIFF
--- a/backend/server/rhnVirtualization.py
+++ b/backend/server/rhnVirtualization.py
@@ -316,10 +316,9 @@ class VirtualizationEventHandler:
         # sysid only.
         #
         # When IdentityType.GUEST, need to worry about cross-org issues...
-        # 3 states to worry about:
+        # 2 states to worry about:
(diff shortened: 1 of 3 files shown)


thedevicesupport:
http://www.rokuhelp.com/

jdobes:
Taking these patches in, the multi-org behavior seems definitely better now.

There are not generated multiple rhnVirtualInstance records with same uuid anymore once registered guest is in another org.

UI patches to display at least some node (with virt host name) when it's in different org is good because without this the list was just broken.

Now the logic seems correct:
1. Host registered in current org => clickable host name, display all guests managed by this host including these not registered in SW
2. Host registered in different org but there is any guest managed by this host in current org => unclickable host name, only guests from current org are displayed

I added 2 more fixes:
ef797939c50eef9676e0355399410e37cbcadaf3, 30b233a6092658dacb4a407d77f2ba678603c9d3

PR #324, opened 2015-12-13
Merge create first org into create new org

mcalmer:
@jdobes : After your changes last week, the first org is created during the creation of the first user.
Java calls the procedure create_new_org() which is a bit different to create_first_org(). 

This commit merge the missing piece from create_first_org() into create_new_org() and drop
the now unused create_first_org() procedure.


jdobes:
Thank you for PR, I missed these procedures. I'm not sure if we need to insert "Satellite Administrators" into the rhnUserGroup table because it seems the row is created anyway (probably somewhere from java code)

```
select name, current_members, group_type, org_id from rhnUserGroup where group_type = 1;
           name           | current_members | group_type | org_id 
--------------------------+-----------------+------------+--------
 Spacewalk Administrators |               1 |          1 |      1
(1 row)
```

If I apply your patch, I'll get one row too, basically the same, just with different texts. Can we skip adding the "if new_org_id = 1" part or I'm missing something?

mcalmer:
Right. I just wanted to be on the safe site and integrated this functionality into create_new_org.
Maybe somebody wants to call it later not from java and is wondering why some things are not working.

mcalmer:
One example might be running java unit tests. They expect that org 1 exists. So somebody need to 
create it before calling the unit tests.

mcalmer:
Seems that only dropping the function is currently missing in the code.

jdobes:

jdobes:
Rebased to master. Thank you.

PR #325, opened 2015-12-13
List global available CryptoKeys

mcalmer:
@jdobes - Global available kypto keys are now added with org = NULL.
This commot make KickstartLister.cryptoKeysInOrg() aware of this.


jdobes:
Actually, there is a bug - I expected the first created organization will copy the RHN-ORG-TRUSTED-SSL-CERT key from null org and then continue as before. Unfortunately, the key stays in null org and nobody can see him.

I fixed it in d84dbd7fa745c557db24733875f5dbd37d34d3de and I think it works as expected now, could you check it?

mcalmer:
@jdobes I have tested your fix in SUSE Manager and it seems to work.
Was a little bit tricky to understand that by requesting the satellite org before first or is created
return NULL and later code copy than from org NULL to the first org, but it works :-)

Close without merge

PR #326, opened 2015-12-15
Fix removing virtual guests

hustodemon:
This aim of this PR is to clean some mess around the `Server <---> VirtualInstance` association:
- cc5c141: `VirtualInstance.hashCode` & `equals` relied on an ID of the entities, which is problematic when the `VirtualInstance` entity also part of some collection (which is true - it has many to one association with the `Server` entity). See [hibernate manual](https://docs.jboss.org/hibernate/stable/core.old/reference/en/html/persistent-classes-equalshashcode.html) for further info.
- 3ed03f5: Fixing removing associations between virtual servers and their guests via the UI
- Rest: Simplifying the deletion of the association between virtual servers and their guests. Previously more methods were used to achieve this, now `VirtualInstanceFactory.deleteVirtualInstanceOnly` is used.


thedevicesupport:
http://www.rokuhelp.com/

jdobes:
Seems working as expected, rebased into master. Thanks.

PR #327, opened 2015-12-15
Change the virtualization backend not to duplicate data in case host and guests are in different organizations

hustodemon:


jdobes:
Isn't this commit already included in #322 ?

hustodemon:
You're right! I'm closing this PR.

PR #328, opened 2015-12-15
fix %tomcat in spacewalk-branding.spec for EL7

Thubo:
Analog to spacewalk-java.spec

### DIFF
--- a/branding/spacewalk-branding.spec
+++ b/branding/spacewalk-branding.spec
@@ -1,7 +1,7 @@
 %if  0%{?rhel} && 0%{?rhel} < 6
 %global tomcat tomcat5
 %else
-%if 0%{?fedora}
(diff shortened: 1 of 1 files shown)


jdobes:
Merged, thanks. Do you know any bugs it was causing?

Thubo:
Not really a bug a far as I know.
However if you install the spacewalk-branding package, using the incorrect .spec file (with %tomcat = 'tomcat6') on an EL7 system Tomcat will refuse to start with the following error message:

```Could not initialize class com.redhat.rhn.common.localization.LocalizationService```

This is due to the symlink to `/usr/share/rhn/lib/java-branding.jar` being in  `/var/lib/tomcat6/webapps/rhn/WEB-INF/lib` and not in `/var/lib/tomcat/webapps/rhn/WEB-INF/lib`.

PR #329, opened 2015-12-16
Set HTTP status code for error pages

renner:
This patch solves the following problems around the JSP error pages:

- Do not require the user to be logged in to see error pages
- Allow error pages to be requested unencrypted via HTTP
- Set the HTTP status code correctly for error pages (404, 500)


ggainey:
Accepted as 323da3ca ab86e567 and dec5b2ae - thanks for the patch!

PR #332, opened 2015-12-23
MessageQueue/ActionExecutor: use generics

dmacvicar:
Some refactoring to use generics


ggainey:
Accepted as commit 94406f42 - thanks for the patch!

PR #333, opened 2015-12-25
Update oracle channels

Djelibeybi:
This fixes a typo introduced in a previous commit, but doesn't add channels we haven't actually published yet. :)


ggainey:
Looks like the net of these three commits was done independently as 68189d56 - I'm closing this PR. Reopen if I missed something!

PR #334, opened 2015-12-28
Osad multiple servers

aleksey-novikov:
These changes allow osad use multiple servers looping throgh server list and search one available for connection.


galtsib:
I've fixed this problem in scope of this bug:

https://bugzilla.redhat.com/show_bug.cgi?id=1306541

PR #335, opened 2016-01-05
Add delimiter option for spacewalk-manage-channel-lifecycle

mbrookhuis:
Customer sometimes need a different delimiter (eg an `_`) the see a better differentiation between the prefix and the original name. Eg the prefix is `dev-dmz`. With the current delimiter a channel would have the name `dev-dmz-sles12-sp1-pool-x86_64`. Customer wants to see `dev-dmz_sles12-sp1-pool-x86_64`.
To achieve this, introduced the parameter `DELIMETER` or `-D`. Here you can choose between `-` or `_`. When the parameter is not being given, `-` will be used. The delimiter will be used between the prefix and the channelname.

### DIFF
--- a/utils/spacewalk-manage-channel-lifecycle
+++ b/utils/spacewalk-manage-channel-lifecycle
@@ -306,12 +306,13 @@ def get_current_phase(source):
 
 
 def build_channel_labels(source):
+    destination = None
     if options.archive:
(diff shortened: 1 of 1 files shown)


ggainey:
Accepted as 918502c65 - thanks for the patch!  

Since channel-name and -label also allows '.' , I added that to the list of accepted delimiters with a3a3053b

PR #337, opened 2016-01-08
fix pyGIWarning: GUdev was imported without specifying a version first.

Pajinek:
/usr/share/rhn/up2date_client/hardware_gudev.py:18: PyGIWarning: GUdev was imported without specifying a version first. Use gi.require_version('GUdev', '1.0') before import to ensure that the right version gets loaded.
  from gi.repository import GUdev


tomas-kasparek:
Committed as 361efdfda34b381157b11bde1e7fcd06c64347c9. Thank you

PR #342, opened 2016-01-20
Log error in AbstractDatabaseAction

mateiw:
Log the exception instead of doing a `printStackTrace()`.


ggainey:
Accepted as commit 2bcce00c - thanks for the patch!

PR #344, opened 2016-01-22
python <2.7 fix 'TypeError: encode() takes no keyword arguments'

Pajinek:
http://docs.python.org/library/stdtypes.html#str.decode

doc: str.encode([encoding[, errors]])
Changed in version 2.7: Support for keyword arguments added, but it isn't used in Python 2.4/2.6


ggainey:
Accepted as commits 7641f361 and  4af86152 - thanks for the patch!

PR #345, opened 2016-01-25
Add Chile to the list of timezones

dmacvicar:
Chile recently switched to follow UTC-03:00 but without DST so there is no compatible timezone to be selected.

Also, the current timezone sorting algorithm only uses the offset, so tweak it to avoid Chile going to the top.

### DIFF
--- a/java/code/src/com/redhat/rhn/domain/user/UserFactory.java
+++ b/java/code/src/com/redhat/rhn/domain/user/UserFactory.java
@@ -540,32 +540,26 @@ public  class UserFactory extends HibernateFactory {
                     public int compare(Object o1, Object o2) {
                         RhnTimeZone t1 = (RhnTimeZone) o1;
                         RhnTimeZone t2 = (RhnTimeZone) o2;
-                        int offSet1 = t1.getTimeZone().getRawOffset();
-                        int offSet2 = t2.getTimeZone().getRawOffset();
+                        Integer offSet1 = t1.getTimeZone().getRawOffset();
+                        Integer offSet2 = t2.getTimeZone().getRawOffset();
 
-                        if (offSet2 - offSet1 == 0) {
-                            return t2.getOlsonName().compareTo(t1.getOlsonName());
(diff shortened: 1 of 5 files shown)


java/code/src/com/redhat/rhn/domain/user/UserFactory.java
```diff
@@ -548,16 +548,20 @@ public int compare(Object o1, Object o2) {
                         }
 
                         if (offSet1 <= -18000000 && offSet1 >= -36000000 &&
-                                offSet2 <= -18000000 && offSet2 >= -36000000) {
+                                offSet2 <= -18000000 && offSet2 >= -36000000 &&
+                                t1.getOlsonName().startsWith("America") &&
```
ggainey: These heuristics end up switching Hawaii and Midway. Canada(Regina) is already in the middle of the 'United States' set. I'm not sure this little piece of US-centrism is worth the fragility. Maybe modify the comparator so that it puts GMT first and then just lists everything east-to-west from there?

dmacvicar:
That is an option. Another is to sort by name, but the current string displayed is not the one that the users will probably look for. What do you think about just showing the Olson name there and sort by that?

ggainey:
The advantage the current setup has, is that the user can choose an appropriate TZ even if they're not clear on which Olsen-name applies to them. And the sort-in-TZ-order makes sense, in that context.

If we want to use Olsen names, we should prob think about using the complete set. We actually *have* a pretty good list already, in rhnKickstartTimezone. Why we have two list/ways of describing timezones, is a question to be answered by some software archaeologist of the future...

dmacvicar:
Ok, first idea of sorting GMT first, then East to West implemented.

ggainey:
Accepted as e3484eaf 2f20508e 95d88a55 9897e47e 38ab9aa0 - thanks for the work!

I tweaked the order with b8570146 - list is now ordered E-to-W starting at GMT

PR #346, opened 2016-01-26
Fix option names to correspond with rhn_server.conf

Klaas-:
I think this is a simple typo - that config option has another name everywhere else :)


jdobes:
Thank. Accepted as efe516272a827fe0d9596381398960ef081c65bc .

PR #347, opened 2016-02-03
Remove custom latest RPM handling in spacewalk-repo-sync

Djelibeybi:
The custom logic was failing for some RPMs in the Oracle Linux repositories. I refactored `yum_src.py` to use the same yum functions as the `reposync` tool from `yum-utils` does.

I also added in the sort function so that the RPMs are downloaded in alphabetical order, like `reposync` does too.


jnovack:
Thanks for this.  It allows me to install remi-php56more.

Hacked in for now, until official rpms are updated.

PR #351, opened 2016-02-15
modified rhn-custom-info to work in python 2/3

Pajinek:
* modified function print, except
* removed module string and method has_key
* update spec file for Fedora 23

### DIFF
--- a/client/tools/rhncustominfo/rhn-custom-info.py
+++ b/client/tools/rhncustominfo/rhn-custom-info.py
@@ -17,7 +17,6 @@
 
 import sys
 import os
-import string
 import getpass
 
 try:
--- a/client/tools/rhncustominfo/rhn-custom-info.py
+++ b/client/tools/rhncustominfo/rhn-custom-info.py
@@ -23,7 +23,7 @@ try:
     import xmlrpclib
 except ImportError:
     import xmlrpc.client as xmlrpclib
-    basestring = str
+    basestring = (str, bytes)
 
 from re import search
 from optparse import OptionParser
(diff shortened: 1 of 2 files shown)


client/tools/rhncustominfo/rhn-custom-info.py
```diff
@@ -17,9 +17,13 @@
...
 import os
-import string
 import getpass
-import xmlrpclib
+
+try:
+    import xmlrpclib
+except ImportError:
+    import xmlrpc.client as xmlrpclib
+    basestring = str
```
galtsib: I believe we can just use 'str' instead of 'basestring' in all places and remove this line.

client/tools/rhncustominfo/rhn-custom-info.py
```diff
@@ -55,13 +59,12 @@ def create_server_obj(server_url):
 
     lang = None
     for env in 'LANGUAGE', 'LC_ALL', 'LC_MESSAGES', 'LANG':
-        if os.environ.has_key(env):
-            if not os.environ[env]:
-                # sometimes unset
-                continue
-            lang = string.split(os.environ[env], ':')[0]
-            lang = string.split(lang, '.')[0]
+        if os.environ.get(env):
```
galtsib: if env in os.environ:

client/tools/rhncustominfo/rhn-custom-info.py
galtsib: for Python 3: basestring = (str, bytes)

Pajinek: I fixed it.

galtsib:
merged

PR #352, opened 2016-02-18
Python24rhel5

Pajinek:
The registration system doesn't work on RHEL5 due to python 2.4 syntax error. I created some litle fixes. And I found out next one problem, that is described in this comment https://github.com/spacewalkproject/spacewalk/commit/d726a1eac0193bed1b36d84f96b2e54bf9af8d76


jdobes:
Thanks.

PR #353, opened 2016-02-25
cut string after full characters

hustodemon:
Sometimes the function for sanitizing strings cuts a UTF character "in the middle" , partially reproducible e.g. when syncing a repo where rpms have too long changelogs.


jdobes:
Adding related reproducer:
https://www.redhat.com/archives/spacewalk-list/2016-February/msg00087.html

jdobes:
Thanks for the patch, it works but I fixed it with simplier patch:

44c7e8b0cf910560f73177ffe65a695290ae1e2f

hustodemon:
wonderful, thanks!

PR #355, opened 2016-03-01
RPC debugging information in satellite-sync

PaulWay:
Pass the debug level into the RPC library's server handler for debug levels above 1 (which is 'normal' in satellite-sync but emits data in rpclib).

### DIFF
--- a/backend/satellite_tools/connection.py
+++ b/backend/satellite_tools/connection.py
@@ -100,8 +100,12 @@ class StreamConnection(_Server):
 
     def __init__(self, uri, proxy=None, username=None, password=None,
                  refreshCallback=None, xml_dump_version=constants.PROTOCOL_VERSION,
-                 timeout=None):
-        _Server.__init__(self, uri, proxy=proxy, username=username,
+                 timeout=None, debug=0):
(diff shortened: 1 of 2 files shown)


PaulWay:
Odd - that Travis build failed because of a GCC package conflict, not because of anything to do with my code.  Can someone have a look at that?

jdobes:
Paul, dynamically change debug levels in tool because some modules use different levels of messages doesn't seem right to me. Is more debug info on level 1 really a problem?

PaulWay:
Hi Jan,

Yes, the problem with leaving it at level 1 is that satellite-sync normally operates at level 1 unless given -d 0 which turns all messages off.  However, level 1 debugging information in rpclib emits the request and response headers, which is not needed for normal operation of satellite-sync.  I discovered this the hard way in my initial testing :-)  Feel free to try it out to see what I mean.

Hope this helps,

Paul

jdobes:

eherget:
Thank you for your contribution.  However we are closing this PR as changing debug level to zero when user expected 1 doesn't seem like the right way to address this.

PR #356, opened 2016-03-02
1232596 - still require openscap-utils on RHEL5

Klaas-:
cherry picked from commit db3287fe19df204a3c9841eb915bdce8ecf37da6

I'm not sure if this is the right way to cherry pick a commit into a release version :) Currently the spacewalk-oscap package is broken on rhel5 because it requires a non-existent package. It's been fixed in 2.5 but not in the current release. I think it should be done to have an actual working stable package for rhel5.

Greetings
Klaas

### DIFF
--- a/client/rhel/spacewalk-oscap/spacewalk-oscap.spec
+++ b/client/rhel/spacewalk-oscap/spacewalk-oscap.spec
@@ -12,7 +12,11 @@ BuildArch:	noarch
 BuildRequires:	python-devel
 BuildRequires:	rhnlib
 BuildRequires:  libxslt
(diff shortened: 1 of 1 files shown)


Klaas-:
@jdobes I see you've merged the commit into the 2.4 branch - is that going to make it into packages on http://yum.spacewalkproject.org/? Do you need someone to test it?

jdobes:
Yes, it will eventually get into the repository, the packages have to be signed yet. However, you can test unsigned build from Spacewalk Koji: http://koji.spacewalkproject.org/koji/buildinfo?buildID=42986

Klaas-:
Okay, thanks for the information - Closing PR

PR #357, opened 2016-03-07
Rhn hidden taglib provide

mcalmer:
rhn:hidden need to handle id as well.
This should fix buttons on the provisioning schedule page.

complete commit 63e3ccec4130622059e9f89fd96bcc83269383d1

@ggainey please check


ggainey:
Oh, good catch - I seem to have lost this part of adding 'id' in 63e3ccec :(

Thanks, accepted as 6547ef2b and 24c44bab !

PR #358, opened 2016-03-09
Fix alignment of immutable entitlements in the UI

renner:
This patch makes them show up aligned with the rest of the form even if immutable (`permanent`).


jdobes:
Looks better. Merged as c9f5023786071ff94461630ab2c4ea02a57e52ee . Thanks.

PR #360, opened 2016-03-15
Fix case statements to correctly check for NULL

renner:
It doesn't seem to work as expected if `CASE x WHEN NULL` is used, instead it apparently needs to be written as `CASE WHEN x IS NULL`.


jdobes:
Right, "WHEN NULL" is always evaluated as false. Merged as b6334f605692dba5fb6567ada768d6cdebba1399 . Thanks.

PR #361, opened 2016-03-17
Disable changing Managers for Vendor Channels

mcalmer:
This UI editable makes only sense for custom channels. For Vendor Channels a readonly
variant is sufficient.

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/action/channels/ManagersSetupAction.java
+++ b/java/code/src/com/redhat/rhn/frontend/action/channels/ManagersSetupAction.java
@@ -64,6 +64,7 @@ public class ManagersSetupAction extends RhnAction implements Listable {
         }
 
         request.setAttribute("channel_name", currentChan.getName());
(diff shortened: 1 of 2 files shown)


jdobes:
Makes sense but could you please move the "<hr />" tag under if in jsp to make it look better? Or are you OK if I do it?

What do you think about not showing "Managers" bookmark at all for vendor channels?

mcalmer:
> What do you think about not showing "Managers" bookmark at all for vendor channels?

I wonder why it is there at all. If we remove it for Vendor channels, than we have the same 
thing as below "Manage Software Channels" which shows only Vendor Channels.

So I think there might be a reason why we have it also in the general overview even if 
I do not know what reason it could be :-)

jdobes:
I'm not fully sure about this feature either, so nevermind.

Merged as one commit - 98e0e6e4c5ef3cc51b2b9d486032585385257911 . Thank you.

PR #363, opened 2016-03-18
Fix for bz1309337 'rhnreg_ks doesn't work with activation key'

BlackSmith:
The fix of bug [https://bugzilla.redhat.com/show_bug.cgi?id=1309337](https://bugzilla.redhat.com/show_bug.cgi?id=1309337).


jdobes:
Merged as 3ba4d5bc30f337f4cba25d0a984e94ffa45f2568 . Thanks.

PR #364, opened 2016-03-18
Fix the string representation of PackageEvr

renner:
This patch fixes the `toString()` method of `PackageEvr` to return a string representation that makes more sense. Also the new format looks more like what the other code actually [seems to expect] (https://github.com/spacewalkproject/spacewalk/blob/master/java/code/src/com/redhat/rhn/domain/rhnpackage/Package.java#L473).


tlestach:
The patch looks good, accepting as a3c9149052b5f58ba854d5c3a74697e535385769.

PR #366, opened 2016-03-21
fix scheduling an action chain

mcalmer:
The rhn:hidden taglib cannot be used here.


tkasparek:
Committed into Spacewalk as f588b874c6d868b27f61cc8a4a84f5bac91fc2ac. Thank you!

PR #367, opened 2016-03-23
Fix: add a missing url mapping

ncounter:
This PR fix the correct `sidenav` highlighted label for the `rhn/kickstart/tree/EditVariables.do` URL since the mapping was missing.

Before:
before

  
After:
after


jdobes:
Thank you.

PR #369, opened 2016-03-31
fix splitting kernel options

mcalmer:
limit -- the result threshold which means how many strings to be returned.

option.split("=", 1) returns the whole string without splitting, so we need to provide "2" to split one time.


tkasparek:
Committed into Spacewalk as 77fd54b47b1732bb68c329527995a25d1fb020da. Thank you!

jdobes:
Nice catch, I was looking at the same code some time ago when I fixed this in koan but I didn't realize it has to be two in Java.

PR #372, opened 2016-04-15
Refactoring reposync

dincamihai:
My goal here is to integrate SUSE/RepoSync class with spacewalkproject/RepoSync so that we can contribute more to this open-source project

Summary:
- I have promoted reposync.py file to a module
- I have added SUSE's version of RepoSync
- I created a parent class used in both versions of RepoSync with the methods that are already common

Note:
I know it's a big diff but I've tried to have single-step commits in order to be as easy as possible to follow the refactoring.

TODO:
- better integration of the two RepoSync versions
- migrate the reposync tests we have at SUSE to this repo (we use nosetests and we run the tests in docker)


jdobes:
Just an idea about different approach. What about putting this vendor specific version of reposync into separate spacewalk-backend subpackage? For example spacewalk-backend-suse, this package would be fully optional to install.

dincamihai:
@jdobes thank you for taking a look and for your suggestion.
This vendor separation into something that it is completely optional would be really nice but I don't consider that to be a completely different approach but maybe something that can improve this PR.

In this PR a have extracted the common code of [RepoSync] and [SuseRepoSync] into a parent class called [BaseRepoSync]. That means [SuseRepoSync] is not completely isolated from the rest of the code. It is actually [RepoSync] with some particularities.
In [backend/satellite_tools/reposync/__init__.py](https://github.com/dincamihai/spacewalk/blob/refactoring-reposync/backend/satellite_tools/reposync/__init__.py) I decide based on **AS_SUSE** environment variable what version of the class to use.
I could replace that with something like:

```python
if os.environ.get('AS_SUSE', False):
    from suse_reposync import (
        RepoSync, ChannelException, ChannelTimeoutException
    )
else:
    from reposync import RepoSync
```

That would allow me to move [backend/satellite_tools/reposync/suse] module to a separate python package and when spacewalk is installed on a system that has the *suse_reposync* package, it would use the custom version. SUSE could also add this custom module as a sub-package for spacewalk.
One thing to mention here is that this would not be possible without the refactoring in this PR.
What do you think?
If you are ok with that, I will remove the [backend/satellite_tools/reposync/suse] module from this PR and I will put it in a separate python package for now.

[RepoSync]: https://github.com/dincamihai/spacewalk/blob/refactoring-reposync/backend/satellite_tools/reposync/reposync.py#L20

[SuseRepoSync]: https://github.com/dincamihai/spacewalk/blob/refactoring-reposync/backend/satellite_tools/reposync/suse/reposync.py#L34

[BaseRepoSync]: https://github.com/dincamihai/spacewalk/blob/refactoring-reposync/backend/satellite_tools/reposync/common.py#L101-L232

[backend/satellite_tools/reposync/suse]: https://github.com/dincamihai/spacewalk/tree/refactoring-reposync/backend/satellite_tools/reposync/suse

eherget:
Thank you for your contribution.  We are closing this PR because there have been many changes to reposync since this PR was submitted.  If you like, please submit a new PR against the latest nightly codebase.

PR #373, opened 2016-04-18
Fix: Taskomatic periodically crashes, terminated with SIGKILL when JVM appears hung

dmacvicar:
Taskomatic uses tanukiwrapper. Tanukiwrapper is a C process that spawns the JVM and does the following extra actions:

* Forward jvm output to the logfile
* Ping the jvm, via a protocol/socket implemented in the `WrapperListener` interface which forces the java daemon to be implemented using a event loop architecture where all tasks happen in threads started from the event loop.

However, there are various problems:

* TaskomaticDaemon checks for tomcat being up inside the event loop, which makes Tanuki think the jvm is stuck. Problem is that after trying a few times, tanuki will leave taskomatic in dead state.
* There is no need to check for tomcat because Taskomatic depends on cobbler API that it is routed via apache and not tomcat (not sure the reasoning behind 6fdefd2bff03a67b04d76c0f7e3cd667434db410). We can express this better using systemd dependencies.
* Under high load, tanuki thinks the JVM is dead when it is not, and kills it. This can be reproduced simulating high load.

So this patchset:

* Removes the check for tomcat at startup of taskomatic
* Disables the ping timeout for Tanukiwrapper
* The last commit 	59d3949 prevents that if a database restart or similar prevents quartz of initializing correctly, that you dont end with a zombie taskomatic that looks running from tanuki point of view but it is not scheduling anything (eg. when SchedulerKernel throws a PostgreSQL exception).

We are also working in getting completely rid of Tanuki and have systemd manage taskomatic directly. You will see a PR into that direction in the future, but this patchset should make the situation better meanwhile.

### DIFF
--- a/java/code/src/com/redhat/rhn/taskomatic/core/BaseDaemon.java
+++ b/java/code/src/com/redhat/rhn/taskomatic/core/BaseDaemon.java
@@ -37,6 +37,7 @@ public abstract class BaseDaemon implements WrapperListener {
     public static final int LOG_INFO = WrapperManager.WRAPPER_LOG_LEVEL_INFO;
     public static final int LOG_ERROR = WrapperManager.WRAPPER_LOG_LEVEL_ERROR;
     public static final int LOG_FATAL = WrapperManager.WRAPPER_LOG_LEVEL_FATAL;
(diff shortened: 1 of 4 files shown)


tkasparek:
Pushed into Spacewalk. Thank you.

PR #374, opened 2016-04-22
Osad python3

Pajinek:
work on code refactoring to work in Python 2/3


client/tools/osad/src/jabber_lib.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2008--2014 Red Hat, Inc.
+# Copyright (c) 2/08--2014 Red Hat, Inc.
```
galtsib: I think it should be better to leave 0 here, but idea is not bad :)

galtsib:
Merged

PR #381, opened 2016-05-01
osad: fix permissions on directories

mcalmer:
Align directory permissions


ggainey:
Merged as b3d3d25a - thanks!

PR #382, opened 2016-05-01
pwstrength-bootstrap: build on openSUSE

mcalmer:


ggainey:
Merged as 4041fec8 - thanks!

PR #385, opened 2016-05-01
select2-bootstrap-css: build on openSUSE

mcalmer:


ggainey:
Merged as 56d1cf20 - thanks!

PR #388, opened 2016-05-01
spacewalk-admin: build on openSUSE

mcalmer:
- openSUSE uses systemd too
- workaround different names httpd.service vs. apache2.service
- workaround different paths for lsof
- workaround different paths for tomcat.service file


ggainey:
Merged as e5570cb1 - thanks!

PR #389, opened 2016-05-01
spacewalk-backend: build on openSUSE

mcalmer:

apache documentroot, user and group is different on openSUSE. make config and code configurable and let it detect if it is running on SUSE or not.


ggainey:
Made a few tweaks to fix RHEL, and then merged as 9050d387 - thanks!

PR #390, opened 2016-05-01
spacewalk-branding: build on openSUSE

mcalmer:

configure tomcat and apache path


ggainey:
Merged as 173d608f - thanks!

PR #392, opened 2016-05-01
spacewalk-client-cert: build on openSUSE

mcalmer:
add some directories to the filelist which are required for openSUSE


ggainey:
Merged as 5b042715 - thanks!

PR #393, opened 2016-05-01
spacewalk-config: build on openSUSE

mcalmer:
- use a macro for apache config dir
- use a macro for apache package name
- use a macro for apache group name
- on SUSE enable required modules for apache


ggainey:
Merged as 6c0346b1 - thanks!

PR #394, opened 2016-05-02
fix dropping unused entitlements when multiple orgs are available

mcalmer:
If multiple organizations are available the schema migration failed with syntax errors.


jdobes:
Merged as 07de4865d826aef6f0bd3efdf065c18969e514fc . Thank you.

PR #396, opened 2016-05-04
1332224 - service osad doesn't work with selinux

Pajinek:
I fix problem with avc message

type=AVC msg=audit(1462198745.064:6466): avc:  denied  { write } for  pid=20175 comm="osad" name="urandom" dev="devtmpfs" ino=6936 scontext=system_u:system_r:osad_t:s0 tcontext=system_u:object_r:urandom_device_t:s0 tclass=chr_file permissive=0


galtsib:
Merged

PR #397, opened 2016-05-09
Don't modify request map when rendering alphabar

hustodemon:
..., since it may fail depending on the implementation of ServletRequest.

In older tomcats (6, 7), the request map was implemented as a HashMap, which survived the modification. This is not true in tomcat 8 and the modification can fail.

Moreover, the documentation of ServletRequest interface says this map is
immutable which suggests that modifying it is a bad idea.

Although this is not reproducible on Spacewalk 2.5.49 with tomcat 7.0.59, this will likely become a problem when tomcat is upgraded to 8 and shouldn't be forgotten.


ggainey:
Accepted as fe8e06cf - thanks!

hustodemon:
you're welcome!

PR #402, opened 2016-05-14
remove monitoring from SUSE spacewalk target

mcalmer:
Monitoring was dropped a long time ago. By accident it was part of the SUSE version of the target.


ggainey:
Accepted as f14bf5fb - thanks!

PR #404, opened 2016-05-15
fix isSUSE check

mcalmer:
Fix the check in isSUSE()


ggainey:
Accepted as ede58fd2 - thanks!

PR #405, opened 2016-05-15
create the symlink directly and point to correct destination on SUSE

mcalmer:
The certificates are created at a different destination on a SUSE system.
This commit creates the symlinks directly in the spec and choose the correct destination on SUSE.


tkasparek:
Pushed into Spacewalk master as f1e5473f2c8b6950d25ad4fca863b8205b31a52a.
Thank you!

PR #406, opened 2016-05-15
spacewalk-oscap: build on openSUSE

mcalmer:
Add missing directories for openSUSE


ggainey:
Accepted as d20fff9a - thanks!

PR #407, opened 2016-05-15
spacewalk-proxy: build on openSUSE

mcalmer:
use macros for apache user and group and add support for different service names


ggainey:
Accepted as 0afba12b - thanks!

PR #410, opened 2016-05-16
convert binary data to string to possible compare data in Python 3

Pajinek:
On line 458 it is compared list of strings with list of binary, it is mistake and rhnpush doesn't work correct in Python 3. It is possible to convert binary to string.

``` if key not in uploadedPackages:
 ...
if ('test-package', '0.2', '2', None) in [b'test-package', b'0.2', b'2', None]
```


tkasparek:
spacewalk.git(master): e18d71c9962e3dfa893a0d71c9280a936bd47e38

PR #411, opened 2016-05-17
Only trigger virtualization notification on server save when the virtualization data is not falsy

hustodemon:
Previously the condition check allowed triggering this event on values such like `''`.


tkasparek:
spacewalk.git(master): 0f12dab08256feb78c726ca27e1e0480a18e3276
Thank you!

hustodemon:
you're welcome!

PR #415, opened 2016-05-20
convert code to work in python 2/3

Pajinek:
Oscap doesn't work in Python 3. I modified code and found problem with call function `xml.sax.parse`, which closed file `f`. It reason why I changed it to xml.sax.parseString.


galtsib:
merged.

PR #416, opened 2016-05-24
call 'queue channel change' only once per channel change

tlestach:
looks like ChannelManager.queueChannelChange is called twice for one
channel change:
 - 1st time within the ChannelManager.refreshWithNewestPackages
 - 2nd time explicitelly
I do not think queueChannelChange needs to be called twice


tkasparek:
Committed into Spacewalk master as 95345b58a86b134084bcc5a8c07ae452f1574670.
Thanks dude.

PR #417, opened 2016-05-24
removing unused code

tlestach:
looks like this method isn't used at all


tkasparek:
Committed into Spacewalk master as 8b2052f68cbed2c4fa0c0600eab15b3380876f39.
Thanks dude.

PR #420, opened 2016-05-26
spacewalk-reports: build on openSUSE

mcalmer:
build on openSUSE


ggainey:
Accepted as aece2aeb - thanks!

PR #425, opened 2016-05-26
spacewalk-setup-postgresql: build and setup on openSUSE

mcalmer:
- runuser does not exist on openSUSE; use su
- lsof is located in /usr/bin
- postgres pid file has a different location


ggainey:
Accepted as f6c3fb60 - thanks!

PR #426, opened 2016-05-28
fix fedora macro usage

mcalmer:
if %fedora is not defined the %if will always evaluate to "true"


tkasparek:
Good catch, committed into Spacewalk master as 2762d21f4355188b94b572565545c73c1376b9e3.
Thank you

PR #427, opened 2016-05-28
fix chgrp call on openSUSE

mcalmer:
apache group on SUSE is "www"


tkasparek:
Makes perfect sense, committed into Spacewalk master as c9217c102ee61513fbe3acb5ad016cd7adb4da7c.
Thank you!

PR #429, opened 2016-06-07
fix bz343653 upload binary file by rhncfg-manager d

Pajinek:
1343653 - Uploading binary file by rhncfg-manager doesn't work in Fedora 23


galtsib:
Merged.

PR #430, opened 2016-06-10
Fix rhncfg verify 

Pajinek:
rhncfg-client causes traceback when try to verify binary file and rhncfg doesn't work correct in python 2.


tkasparek:
Thanks dude

PR #435, opened 2016-07-01
Traceback: comparison method violates its general contract

jiridostal:
Since java 1.7, it's mandatory to deal with Comparator interface specific way, according to [this link](https://docs.oracle.com/javase/7/docs/api/java/util/Comparator.html). It wasn't done like that in Spacewalk, which results in "comparison method violates its general contract". I tried to fix as much occurrences as possible.


ggainey:
SystemSearchHelper needed a lot more help, redone under separate commit 822d97aa
The rest of these changes reviewed and committed as 660362d5

Thanks for the fixes!

PR #437, opened 2016-07-15
keep whitespaces in debian package dependencies. 

hlawatschek:
A different attempt to fix Bug 1226329. In my opinion this tiny fix may be enough to solve the issue. Creation of Packages.gz looks promising.

### DIFF
--- a/backend/server/importlib/debPackage.py
+++ b/backend/server/importlib/debPackage.py
@@ -131,8 +131,7 @@ class debBinaryPackage(headerSource.rpmBinaryPackage):
             l = []
             values = header[k]
             if values is not None:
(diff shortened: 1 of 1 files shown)


ggainey:
Accepted as 153f3fcd - thanks for the patch!

rpasche:
This looks quite good right now. I just tested this with a smaller channel and deleted all packages before. Resynced and Packages.gz is now parseable :-) Although there are now 2 spaces between the operator and the version number. But this is complaining on a very high level.

Thanks.

rpasche:
The only thing I currently can complain about is that it looks like you have to completly resync all channels so that the packages get imported **better**. :+1:

PR #438, opened 2016-07-16
change default checksum type to sha256 for debian packages.

hlawatschek:
Usage of SHA256 is recommended in https://wiki.debian.org/RepositoryFormat#Size.2C_MD5sum.2C_SHA1.2C_SHA256.2C_SHA512
This should also fix RH BZ 1348321


ggainey:
Accepted as b06f0113 - thanks for the patch!

PR #440, opened 2016-07-24
1358484 - dest file in diff output prefixed with '+++'

eherget:
simple fix for a problem I noticed working on the binary diff sane output.


tkasparek:
Committed into Spacewalk master: aaecc727439e8077baf08294d6ce744b09e0b6e1

PR #441, opened 2016-07-26
Setsoftwaredetails

ogajduse:
adding new fuction

### DIFF
--- a/spacecmd/src/lib/utils.py
+++ b/spacecmd/src/lib/utils.py
@@ -770,3 +770,9 @@ def file_is_binary(self, path):
     except OSError:
         pass
     return True
+
+
+def string_to_bool(input_string):
+  if type(input_string) != type(True):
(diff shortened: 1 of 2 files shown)


spacecmd/src/lib/utils.py
```diff
@@ -770,3 +770,13 @@ def file_is_binary(self, path):
...
+
+
+def string_to_bool(input_string):
+    if type(input_string) != type(True):
+        if input_string.lower().rstrip(' ') == 'true':
+            return True
+        else:
+            return False
+    else:
+        return input_string
```
tlestach: what about simplifying the string_to_bool function in following way?

def string_to_bool(input_string):
￼ +    if type(input_string) != type(True):
￼ +        return input_string.lower().rstrip(' ') == 'true'
￼ +    return input_string

tkasparek:
Merged

PR #442, opened 2016-08-04
Migrate to jMock2

hustodemon:
Notable advantages:
- possibility to mock classes (not just interfaces)
- no need to deal with the `Mock` object. One can work with the class of the mocked object.
- the jMock2 DSL is slightly less annoying than the one in the ver. 1.

For more info refer to the [jMock2 cookbook](http://www.jmock.org/cookbook.html).

We run the tests with jMock 2.6.0, but they should be runnable with ver. 2.8 (e.g. in newer Fedoras), too.


ggainey:
Need to get jmock2-2.6+ available in the default ivyserver before I can test this

jdobes:

MichaelMraka:

PR #444, opened 2016-08-12
Improve debian package handling

hlawatschek:
This pull request basically allows the import of debian packages with more complex package names and reintroduces the '_' as separator of debian package name and package version (at least for the spacewalk debian client).

Currently the import mechanism of debian packages can only process simple package names and  one '-' in the version field. I.e. the following package name is not imported correctly: libcunit1_2.1-0.dfsg-10_amd64.deb. This pull request tries to remove this restriction. 
Nevertheless, allowing more complex package names requires some additional modifications which are also addressed by this pull request:
Debian packages use '_' to distinguish between package name and package version. Therefore, the packages.gz file needs some modification and the parseRPMFilename function should be able to distinguish between rpm and deb file names.

### DIFF
--- a/backend/common/rhnLib.py
+++ b/backend/common/rhnLib.py
@@ -157,7 +157,7 @@ def hash_object_id(object_id, factor):
     return num_id.rjust(factor, '0')
 
 
-# reg exp for splitting package names.
+# reg exp for splitting rpm package names.
 re_rpmName = re.compile("^(.*)-([^-]*)-([^-]*)$")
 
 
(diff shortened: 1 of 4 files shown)


jdobes:

wraiden:
The regex in backend/common/rhnLib.py and the import version handling in backend/common/rhn_deb.py is not right for the upstream_version and debian_revision fields.
I'm working on a fix for the debian handling in spacewalk right now taking into account all the information i found in the documentation here: https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Version 
I have already added a propper epoch field handling and fixed the client version reporting so that everything matches on server side for patch management functionality.
I prepare a github repository for this.
I will post another comment here when i have the fixed version handling on server side ready.

I also created client packages for Ubuntu 16.04 based on current spacewalk 2.6 sources, but i will need some help in getting the dependencies right and to get everything in a good shape for debian...

But currently spacewalk does not import the Multi-Arch and some other package header fields into the database and will not write the fields out to its Package.gz catalog. This blocks the usage of spacewalk as a repository-service for debian and ubuntu right now. I will use spacewalk just for patch management and fetch the packages from upstream mirrors on client side as long as this is not fixed (I'm willing to take a look into this but it will involve a database schema evolution and currently i have no clue about spacewalks database infrastructure... who wants to help?).

wraiden:
Please take a look at this:
https://github.com/wraiden/spacewalk/commit/092d5d98b03f1b5825921e02a24b60a2555280b7

wraiden:
added #500 for this...

eherget:
Thank you for the contribution.  I believe the recently merged PR #500 accomplishes what this PR does, and more.  Closing this PR.

PR #445, opened 2016-08-14
fix rhnpush using package_from_filename() function

mcalmer:
changes 02be9116b3954b5448c45b46db742fbe745efaec

rhnpush uses multiple times package_from_filename() in old style.
I think it is easier to append the new parameter to the function and make it optional.
This should fix also rhnpush.

### DIFF
--- a/backend/common/rhn_pkg.py
+++ b/backend/common/rhn_pkg.py
@@ -56,13 +56,15 @@ def package_from_stream(stream, packaging):
     return a_pkg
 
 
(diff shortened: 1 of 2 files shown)


jdobes:
I decided to not change package_from_filename header in 2888ee2d57c9cfcd2c01ff48f7b7b2f35b7a8409 because I can close that stream from outside. (I hope) But thanks for finding this bug anyway, I totally missed rhnpush dependency.

PR #446, opened 2016-08-17
Make heading match navigation: "Managed Systems"

renner:
This very small patch is just supposed to make the page heading match the link text in the left navigation, this is the affected page URL:

- https://(server)/rhn/configuration/system/ManagedSystems.do

Let me know if you want me to touch the translations as well, I could at least adapt the `<source>` tags for all languages if needed.


ggainey:
RE translations - at this point we should only be editing the en_US versions in git, as that is the 'source' for Zanata and the community translators. (https://fedora.zanata.org/project/view/spacewalk-frontend?dswid=-4554 )

ggainey:
Accepted as commit b57c3baa - thanks for the patch!

PR #449, opened 2016-08-24
Avoid a deadlock when deleting a server

moio:
This fixes deadlock condition between update_needed_cache and delete_server.

## Root cause

`rhn_server.update_needed_cache()` explicitly locks `rhnServer` table, then `DELETE`s from `rhnServerNeededCache` (implicitly locking it).
`delete_server()` `DELETE`s from `rhnServerNeededCache` first and then `DELETE`s from `rhnServer` (implicitly locking it).

The difference in locking order gives rise to the deadlock.

## Example log trace

```
2016-01-20 12:50:47 MST susemanager susemanager ERROR:  deadlock detected [...]
	Process 31832: SELECT delete_server(1000010481)
	Process 17164: select * from rhn_server.update_needed_cache($1) as result
```

## Reproducer

To reproduce:
 - prepare a server and register a client
 - run the following snippet on the server to continuously force errata cache recomputation:
```bash
while true; do spacewalk-sql - <<<"select rhn_server.update_needed_cache(id) from rhnserver;"; echo "Updated"; done
```
 - run the following snippet on the client to continuously re-register:
```bash
while true; do ./bootstrap.sh ; done
```
 - watch `/var/lib/pgsql/data/pg_log/postgresql-*` for deadlocks. In my setup I could obtain one every few minutes.

## Solution

In `delete_server()`, explicitly lock `rhnServer` first. That makes the ordering between the two procedures consistent and deadlock cannot happen any more.

After applying the change the bug is not reproducible any more.


ggainey:
Another deadlock bites the dust...

Accepted as c300f76f - thanks for the commit!

PR #451, opened 2016-09-02
Show user a meaningful page after requesting details of non-existing Action Chain

hustodemon:
Previously the page threw Internal Server Error.

* Before
before

* After
after


ggainey:
Accepted as 505c47f3 - thanks for the patch!

PR #456, opened 2016-09-12
Fixes in spacewalk{,-common,-oracle,-postgresql,-base-minimal} package descriptions

lhellebr:
Removed mentions of Solaris.
Mentioned difference for oracle and postgresql backends.
Fixed typos.


jdobes:
Merged into master. Thanks!

PR #457, opened 2016-09-13
  bz#1316708 - added the capability for spacewalk-debug to capture the pg_catalog information from PostgreSQL

tchellomello:
Included pg_catalog.pg_proc on the spacewalk-debug logs which is very useful for troubleshooting


jdobes:
Added, thanks.

63001bc1a3fee2fdae7d06518c02eb0897fa4d90

Updated the condition to not rely on existing path.

9681ee121339555099c2d706c21fa3b92ae052c0

PR #461, opened 2016-09-22
fix bz1340031 - yes_no: command not found

Pajinek:
I run installer of proxy and I see error message in output "/usr/sbin/configure-proxy.sh: line 161: yes_no: command not found".


jdobes:
Commited as 8e3961dd7e931b7d0c134ea50155f05fed6037cd.

Thanks.

PR #466, opened 2016-11-11
Errata schedule notification links

cbbayburt:
This PR adds links to schedule summaries in errata schedule notifications.

When a single patch is scheduled, it should always create a single action. In this case, the link will go to the action details page.

When multiple patches are scheduled, it may or may not create more than one action. In this case, regardless of the action count, the link will point to the pending events list page.


ggainey:
Accepted as 2ebfb531 and 7f5d103a - thanks for the contribution!

PR #467, opened 2016-11-11
Sort proxy clients

cbbayburt:
## Summary
1. Sort the proxy clients list (by name)
2. Have the client names as links to the corresponding system details pages
### Sort the proxy clients list

The simplest way to do this is in the DB query, since this list has a specific query exclusively used here.
### Client names as links to system details

Since the user may not have access to some of the clients in the list, only the ones which the user has access must be shown as links.
### Example

[ActivatedSystemsAction.java@62-89](https://github.com/SUSE/spacewalk/blob/Manager/java/code/src/com/redhat/rhn/frontend/action/token/ActivatedSystemsAction.java#L62-L89)
[activationkeys/systems/list.jsp@40-49](https://github.com/SUSE/spacewalk/blob/Manager/java/code/webapp/WEB-INF/pages/activationkeys/systems/list.jsp#L40-L49)

### DIFF
--- a/java/code/src/com/redhat/rhn/common/db/datasource/xml/System_queries.xml
+++ b/java/code/src/com/redhat/rhn/common/db/datasource/xml/System_queries.xml
@@ -2305,6 +2305,7 @@ SELECT SP.server_id ID, S.name
   FROM rhnServer S, rhnServerPath SP
  WHERE SP.proxy_server_id = :sid
    AND S.id = SP.server_id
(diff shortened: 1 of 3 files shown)


grinrag:
Hello,
Thank You very much for your contribution, but we have already implemented this feature with a bit different way. Almost the same, but also with a possibility to add systems to SSM.

I'm closing this PR. Feel free to open a new one if You have some ideas how to improve our solution.

PR #468, opened 2016-11-12
add channels for openSUSE Leap 42.2

mcalmer:
Update data for spacewalk-common-channels with openSUSE Leap 42.2


jdobes:
We should accept this before SW 2.6, sorry.

PR #472, opened 2016-11-21
Update 'view/modify file' action buttons text

cbbayburt:


eherget:
Thank you for your contribution!  Having the button labels match the names used elsewhere on the page is a good thing!  I added the word "Latest" back to the button labels to make clear that the latest revision will be copied.
This is found in:
spacewalk-java-2.7.66-1

PR #473, opened 2016-11-21
Allow sorting on advisory name in errata lists

cbbayburt:
This PR basically changes three things:

1. Tiny refactoring on relevant-errata-list.jspf
2. Allows sorting on advisory names
3. Refactors ColumnTag code sort direction part, and fixes it so whenever a different column is clicked for sorting, it'll be sorted in ascending order first (previously it was just the opposite of the last sorting direction even though the clicked column is different from the last sorted one - basically switching the sort direction regardless of the change in the sorted column).


ggainey:
Committed as 21f12b4d 649dc435 and 802b917c - thanks for the contribution!

PR #476, opened 2016-11-22
Use different symbols for collapsible sidebar items

moio:
This makes it easier to distinguish menu items visually:
 - use a small circle for items that have no child items
 - use a rightward pointing arrow for items that are collapsed
 - use a downward pointing arrow for items that are uncollapsed

Before:
before

After:
after

### DIFF
--- a/branding/css/spacewalk-theme.less
+++ b/branding/css/spacewalk-theme.less
@@ -267,8 +267,18 @@ aside .nav > li > a {
                 color: @spacewalk-aside-second-submenu-hyperlink-hover;
                 .glow-text;
         }
-	a:before {
+	li a:before {
+		content: "◦";
+		font-family: 'FontAwesome';
+		margin-right: .5em
+	}
+	li.parent a:before {
 		content: "\f105 ";
+		font-family: 'FontAwesome';
+		margin-right: .5em
+	}
+	li.parent.active a:before {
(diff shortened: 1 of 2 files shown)


moio:
Note: this needs Java 8

thedevicesupport:

thedevicesupport:
http://www.rokuhelp.com/

eherget:
Thank you for the contribution here, I think it is a nice enhancement.  There are two changes I'd like to see:
1) we need to support java versions < 1.8, so the java.util.streams should come out.
2) the small 'o' (I think it might be the degree symbol) is just a bit too small.  I'm not sure what to replace it with.  I played with replacing f105 with f0da, f107 with f0d7 and then use f105 in place of the small circle.

eherget:
Thank you for your contribution!  I've pulled it in as-is, and then added a change to support java versions < 1.8 (by removing java.util.stream stuff).  This will appear in:
spacewalk-branding-2.7.4-1
spacewalk-java-2.7.63-1

moio:
Thanks!

eherget:
I just fixed an issue with child nav items not displaying as active.  This was introduced when I did fix to support java versions < 1.8.  This fix appears in:
spacewalk-java-2.7.9-1

PR #479, opened 2016-11-23
Use human-parseable dates for server notes

moio:
Before:
before

After:
after


jdobes:
I looks better. Thanks.

4974b94f321a760bcaaf59a0a76ea4f24dcd66cb

PR #485, opened 2016-11-25
Use consistent naming for SSM pages

moio:
This does the following:
 - change the SSM "main" pages title into "System Set Manager Overview". This is consistent with main pages of other features, eg. "Autoinstallation" -> "Autoinstallation Overview"
 - renames, both in the navigation menu and in the title, the "Status" page to "System Set Manager Task Log". This better reflects the content and is consistent with detail pages of other features, eg "Autoinstallation > Profiles" -> "Autoinstallation Profiles"
 - changes the ordering of tabs in the Task Log page, so that the tab opened by default ("All") is the first. This is also consistent with many other tabbed pages in the Web UI

Overview page example:
overview

Task Log page example:
tasklog

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/strings/jsp/StringResource_en_US.xml
+++ b/java/code/src/com/redhat/rhn/frontend/strings/jsp/StringResource_en_US.xml
@@ -14727,7 +14727,7 @@ any of these revisions.  Are you sure you want to do this?</source>
           <context context-type="sourcefile">/rhn/ssm/ViewLogDetails.do</context>
         </context-group>
         <trans-unit id="ssm.operations.viewlog.header">
-          <source>Tasks Log</source>
+          <source>Task Log</source>
         </trans-unit>
         <trans-unit id="ssm.operations.viewlog.description">
           <source>Description</source>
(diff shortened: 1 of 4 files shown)


eherget:
I'd like to pull in this useful enhancement.  My only concern is changing the name of the System Set Manager page to "System Set Manager Overview" when there is an "Overview" tab on the page in addition to several others.  Also, please make the "Task Log" vs "Tasks Log" consistent.

moio:
> please make the "Task Log" vs "Tasks Log" consistent

Addressed.

> My only concern is changing the name of the System Set Manager page to "System Set Manager Overview" when there is an "Overview" tab on the page in addition to several others

I still find it better than the previous strings - what would you propose instead?

eherget:
> I still find it better than the previous strings - what would you propose instead?

I would just leave it as "System Set Manager" since each tab will have its own subsection title.  It will also match the nav bar text as most of the other nav items do.

eherget:
Thank you for your contribution!  I have pulled this in.  I did revert the System Set Manager page to its original name.  Having Overview in the title is inconsistent with our page titles and and the Overview tab is only 1 of 10 tags on this page.
This change appears in:
spacewalk-java-2.7.64-1

moio:
Thanks!

PR #488, opened 2016-11-29
Differentiate writable/non-writable fields

ncounter:
Give to `<div>` tag for non-writable fields the same aspect as `<input>` tag for `readonly` fields.

Before:
before-readonly

After:
after-readonly


jdobes:
Nice to have. Thanks.

19739b3368a6c3ccdb2856a94fb75ca9ecc3248e

PR #489, opened 2016-12-02
urlsplit back compatibility with python 2.4

alda519:
fixing bug introduced in dc7ee6d

      File "/usr/share/rhn/up2date_client/rpcServer.py", line 75, in _request1
        typ = parse_res.scheme
    exceptions.AttributeError: 'tuple' object has no attribute 'scheme'


jiridostal:
baaf6a971c6c3ddc1c5faf5501c1ddaf0d885fc2

PR #490, opened 2016-12-02
1384093 - RFE: API call to fail scheduled action for the specific systems

ogajduse:
According to BZ1384093 I did some changes. 

Here is the result after calling both variants of this call.
failedsystemspage

### DIFF
--- a/java/code/src/com/redhat/rhn/common/db/datasource/xml/System_queries.xml
+++ b/java/code/src/com/redhat/rhn/common/db/datasource/xml/System_queries.xml
@@ -2299,14 +2299,6 @@ ORDER BY E.update_date, E.id
   </query>
 </mode>
 
-<callable-mode name="fail_action_for_system">
-    <query params="aid, sid, message">
-        UPDATE rhnServerAction S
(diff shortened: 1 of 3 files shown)


jiridostal:
Wouldn't it be possible to use method `setStatus(ActionStatus statusIn)` from package `action.server.ServerAction`? Then we might not need extra SQL query.

java/code/src/com/redhat/rhn/common/db/datasource/xml/System_queries.xml
```diff
@@ -2299,6 +2299,14 @@ ORDER BY E.update_date, E.id
   </query>
 </mode>
 
+<callable-mode name="fail_action_for_system">
```
ogajduse: True, write-mode is more suitable for this, but no longer needed. I have decided to do it another way. See cb43a5584a766504e5dbabb6abe056b32652d38b

ogajduse:
@jiridostal cb43a55 works fine. Can it be considered as the final solution or we should deal with deleting systemactions too?

ogajduse:
thanks to all, implemented in f952fd3adde83cd958b4bff5571228aec89130cc

PR #492, opened 2016-12-06
request repodata regeneration even if an erratum without new packages…

tlestach:
I've found this commit on an older computer. Honestly I'm not sure, whether the code still works or if it's complete. Creating PR to review the patch by the team.


tkasparek:
Hello Tomas,

I've reviewed your patch and it does what's said in commit message. But I have no idea why we want to do this. By any chance is it possible that there was some BZ attached to that commit?

Thank you

tlestach:
Not sure about any associated BZ. But if you sync an erratum without associated packages, you still want the erratum to appear in the repodata, right?

ggainey:
text-only errata, would be my guess...

ggainey:
Accepted as 0503cdea - thanks for the contribution!

PR #494, opened 2016-12-07
python 3 requires print with parentheses

alda519:
fixing sytax error introduced in 8af97691a1

    File /usr/bin/spacewalk-create-channel, line 443
       print fullDir
                   ^
     SyntaxError: Missing parentheses in call to 'print'


jiridostal:
fb8681734241873abf1b5d7e978e3de1da41dfca

PR #495, opened 2016-12-09
dnf-plugin-spacewalk updated to dnf 2.0

MichaelMraka:
this is an update for dnf-plugin-spacewalk to work with dnf-2.0
dnf-2.0 is currently only in rawhide (will be Fedora 26) and the current Fedoras (24 and 25) still use dnf-1.0


fermulator:
I think this might be addressing
https://bugzilla.redhat.com/show_bug.cgi?id=1409320
Yes?

MichaelMraka:
Yes, it seems to be fine with DNF 2.0 plugin.
```
# rhn_check -vv
D: do_call packages.checkNeedUpdate('rhnsd=1',){}
D: local action status: (0, 'rpm database not modified since last update (or package list recently updated)', {})
D: rpcServer: Calling XMLRPC registration.welcome_message
#
```

tkasparek:
Merged into Spacewalk, thank you!

PR #497, opened 2016-12-29
BZ#1401488: Fixes the following error on errata-sync: ERROR: cannot concatenate 'str' and 'int' objects

phurrelmann:


jdobes:
Resolved in https://bugzilla.redhat.com/show_bug.cgi?id=1401488. Thank you.

PR #498, opened 2017-01-05
Added password option for .rhncfgrc

Djelibeybi:
According to https://access.redhat.com/solutions/489553 this feature was being tracked by an internal RFE for Satellite 5.6 back to 2013, but has never seemed to hit the codebase.

This change allows users to add a `password` option to either `/etc/sysconfig/rhn/rhncfg-manager.conf` or `~/.rhncfgrc` under the `[rhncfg-manager]` section, as requested.

### DIFF
--- a/client/tools/rhncfg/config_common/handler_base.py
+++ b/client/tools/rhncfg/config_common/handler_base.py
@@ -62,13 +62,19 @@ class HandlerBase:
             if not username :
                 username=local_config.get('username')
             if not password :
-               (username, password) = self.get_auth_info(username)
+                password=local_config.get('password')
 
             try:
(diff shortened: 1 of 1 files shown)


eherget:
I would like to accept this pull request after two things are addressed.
1) update the rhncfg-manager man page in client/tools/rhncfg/config_management/rhncfg-manager.sgml
2) modify the code to only call login() with username and password args one time - attempt to get password from config, if password variable is still not set, prompt the user.  Then call login(username, password)

eherget:
This has been implemented in PR 543, which included the items requested in a previous comment.  It will be found in rhncfg-5.10.105-1.
Thank you for contributing!

PR #503, opened 2017-02-05
spacecmd: Add a type parameter to repo_create

wraiden:
This adds a type parameter to the repo_create command.
It is needed to add a deb or uln repository from cli.


jdobes:
Hello, this doesn't work in interactive mode:

`spacecmd {SSM:0}> repo_create
ERROR: A name is required`

It's caused by default value:
`Option('-t', '--type', default='yum', action='store'),`

Could you try to fix it, please?

wraiden:
I fixed the failure as requested and implemented the default value for the cli otherwise to keep scripts using the call without the type option working.

Could you also take a quick look over at pull 500 please?
I have no idea whats wrong with the travis check there, maybe you could give me a hint...

jdobes:
Rebased into master. Thanks. Sure, I'll check 500.

PR #504, opened 2017-02-08
Make spacecmd prompt for password when overriding config file user

lrochfort-oracle:
If user credentials are set via spacecmd config file and the spacecmd user is overridden with the -u option, the spacecmd still tries to use the password stored for the user in the config file.

This change makes spacecmd either prompt for a password, or accept one via the -p option.


lrochfort-oracle:
Account compromised. Closing.

PR #505, opened 2017-02-08
Fix spacewalk-data-fsck removing SRPMs associated with RPMs

lrochfort-oracle:
spacewalk-data-fsck incorrectly identifies for removal SRPMs that are associated with RPMs in a channel. If used with the -r option spacewalk-data-fsck will remove the SRPMS.


lrochfort-oracle:
Account compromised. Closing.

PR #506, opened 2017-02-15
fix bz1422518 - request failed: error reading the headers 

Pajinek:
Clients don't download metadata from Satellite due to new httpd security patch:
Bug 1412974 - CVE-2016-8743 httpd: Apache HTTP Request Parsing Whitespace Defects [rhel-6.9]


jdobes:
Rebased into master. Thanks.

pfl202:
Hi, excuse my ignorance, but is it possible to get this fix back ported to SPACEWALK-2.3 client rpms ?

tkasparek:
@pfl202 Hello, I don't think this will get backported into Spacewalk 2.3.
On the other hand using newer client packages with Spacewalk 2.3 should not create any issues.

pfl202:
Thanks for info. I will try latest client.

PR #511, opened 2017-03-01
Avoid "stuck" tabs

moio:
If a page renders slowly for whatever reason, users might open other tabs to visit other pages while it completes.

This is currently blocking - all tabs will not start rendering before the slow one has completed.

Note that this is different from having two distinct browsers - it only applies if the same cookie is used, which maps to the same session and ultimately to the same contended database row.

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/security/AuthenticationService.java
+++ b/java/code/src/com/redhat/rhn/frontend/security/AuthenticationService.java
@@ -46,6 +46,14 @@ public interface AuthenticationService {
         throws ServletException;
 
     /**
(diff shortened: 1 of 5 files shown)


jiridostal:
So far I wasn't able to reproduce the issue - I always succeed to render the other tab while the first one was loading. I've tried it the 'natural' way - search in a large list, as well as doing code changes like putting sleeps in the code to slow server down. I've even tried throttling the connection down. In all cases, I was able to open other tabs with the came cookie. Is your PR still relevant?

moio:
This is 9 month old now, it might well be that something in master has changed in the meantime. Feel free to close this if it the issue is not there any more. Thanks

PR #519, opened 2017-03-19
make country, state/province and city searchable for system location

mcalmer:
System country, state and city could be set, but you cannot search for it.
This PR add these fields to search options. The index exists already.


phillipskimberly:
https://www.thedevicesupport.com/kindle-support/

eherget:
Please revert all the StringResource file changes except the en_US one.  The en_US StringResource file is the default, so when a newly added id is not found in a non-en_US StringResource file, the one found in en_US is used.  This is to avoid issues in the separate process used for translations.

mcalmer:
Ok, the last time somebody forced me to add all the strings in all files to make it easier for translators.
Seems something changed in the process...

eherget:
Thank you for your contribution!  This is merged into spacewalk-java-2.7.78-1

PR #521, opened 2017-03-20
1434037 - Make spacecmd prompt for password when overriding config

lrochfor-oracle:
If user credentials are set via spacecmd config file and the spacecmd user is overridden with the -u option, the spacecmd still tries to use the password stored for the user in the config file.

This PR makes spacecmd prompt for a password when overriding config user.


phillipskimberly:
https://www.thedevicesupport.com/kindle-support/

PR #529, opened 2017-03-31
add new option skip-services-restart

Pajinek:
In some special situation I don't want restart service at end  of installation and I prefer skip this step. For example when spacewalk is installed in Docker.


phillipskimberly:

MichaelMraka:
Thanks for your contribution.

PR #531, opened 2017-04-06
do not fail with a traceback when media.1 is requested

mcalmer:
New libzypp request media.1 directory which result in a traceback.
Just returning 404 Not Found is ok.


MichaelMraka:
Thanks for your contribution.

PR #533, opened 2017-04-07
Enable detection of Oracle Linux during registration.

Djelibeybi:
This patch provides more accurate detection of Oracle Linux clients during registration with Spacewalk.

Detection of Oracle Linux is done first because it provides both `oraclelinux-release` and `redhat-release`.


MichaelMraka:
Thanks for your contribution.

PR #541, opened 2017-05-10
Update system group details page

cbbayburt:
Updates some elements in the system group details page to make it more unified with the rest.


grinrag:
Thank You. Merged.

PR #542, opened 2017-05-16
Make schedule notification links consistent for actions for a single system

cbbayburt:
Instead of pointing to `schedule/ActionDetails.do` all the schedule action notifications for a single system will point to `systems/details/history/Pending.do`


eherget:
Thank you for your contribution.  however, we are closing this as we prefer the current behavior.

PR #543, opened 2017-05-16
Add password config option to rhncfg-manager.

lrochfor-oracle:
According to https://access.redhat.com/solutions/489553 this feature was being tracked by an internal RFE for Satellite 5.6 back to 2013, but has never seemed to hit the codebase.

This change allows users to add a password option to either /etc/sysconfig/rhn/rhncfg-manager.conf or ~/.rhncfgrc under the [rhncfg-manager] section, as requested.


eherget:
Thank you for your contribution!  This will be found in rhncfg-5.10.105-1

PR #544, opened 2017-05-17
Remove redundant 'Channels' word in Software Channels nav item

cbbayburt:


eherget:
Thank you for your contribution.  However, we are closing this as we prefer the labels as-is.

PR #545, opened 2017-05-22
reduce code to normalize path for url

Pajinek:
reduce code for getting normalize url

### DIFF
--- a/backend/cdn_tools/repository.py
+++ b/backend/cdn_tools/repository.py
@@ -391,18 +391,7 @@ class CdnRepositoryTree(object):
     @staticmethod
     def normalize_url(url):
         """Splits repository URL, removes redundant characters and returns list with directory names."""
-        path = []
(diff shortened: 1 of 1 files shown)


tlestach:
I may be off, but the suggested code returns different results than the original one.
E.g.
>>> [part for part in os.path.abspath('http://www.repo.com/with/great/packages').split('/') if part]
['home', 'tlestach', 'http:', 'www.repo.com', 'with', 'great', 'packages']

while the original code returns:
['http:', 'www.repo.com', 'with', 'great', 'packages']

I do not think os.path.abspath is the right thing to use for repository urls.
Rejecting the PR.

Pajinek:
I suppose that correct value is only path. Without protocol and hostname.

comment from code:

`url is relative CDN url - e.g. /content/dist/rhel/server/6/6Server/x86_64/os`

PR #548, opened 2017-06-07
Add a jUnit test for ActionChainManager

moio:
First four commits just add cleanups, the last one adds a new test.


ggainey:
Looks like we're missing com.redhat.rhn.testing.JMockBaseTestCaseWithUser?

ggainey:
We don't have the base class the main thrust of this PR is based on. Closing.

PR #551, opened 2017-06-13
Fix: hide non-org event details

mbologna:
When a server is migrated from an organization to another organization,
its associated actions are tied to the original org.
If user tries to show event details for "migrated" events, will get an
ISE, as:

https://github.com/spacewalkproject/spacewalk/blob/f4c8ad60fdd9dc7b8f40861287be4ab6933a70eb/java/code/src/com/redhat/rhn/frontend/action/systems/sdc/SystemHistoryEventAction.java#L62

will return null.

This patch fixes the problem by not displaying a link to show event
details for all the events that does not belong to the organization
linked to the user who is visiting the page.

Signed-off-by: Michele Bologna <michele.bologna@gmail.com>

### DIFF
--- a/java/code/src/com/redhat/rhn/common/db/datasource/xml/System_queries.xml
+++ b/java/code/src/com/redhat/rhn/common/db/datasource/xml/System_queries.xml
@@ -2011,7 +2011,7 @@ select event_id as id,
        all_events.history_type_name,
        all_events.history_status,
        case when action_details.org_id != :oid then 0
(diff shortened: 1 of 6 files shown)


jiridostal:
Thank you for submitting a pull request. Unfortunately code does not build due to
`com/redhat/rhn/frontend/security/PxtAuthenticationService.java:24: error: package org.apache.http.client.utils does not exist`

PR #552, opened 2017-06-15
Persistence of the current set

ncounter:
Problem: when selecting an errata or a package for a system from the list, reloading the page or going away from it will delete the selection. This is not compliant with the `rhnSet` behavior for *errata_list* (for instance *system_list* is always persistent), but also for *package*s that the user selects and store in the session: the expected behavior should be to have that selection persistent.

Fixes of this patch:
- do not clear `rhnSet` table for `errata` but keep selection persistent loading the list page e502d76fd5eb3310594a49955629ab7491118f95
- do not clear `session set` for `packages` but keep selection persistent loading the list pages b2958b81ef1a86523a784f1a32e9b1bd1871dc82

Pages affected:
- [ ] `/rhn/systems/details/ErrataList.do`
- [x] `/rhn/systems/details/packages/PackageList.do`
- [x] `/rhn/systems/details/packages/UpgradableList.do`
- [x] `/rhn/systems/details/packages/InstallPackages.do`
- [x] `/rhn/systems/details/packages/VerifyPackages.do`
- [x] `/rhn/systems/details/packages/ExtraPackagesList.do`

@grinrag : can you please take a look? Do you have any clue why it was implemented like this lack of persistence with the *clear-on-refresh* mechanism? Thank you in advance.

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/action/rhnpackage/BaseSystemPackagesAction.java
+++ b/java/code/src/com/redhat/rhn/frontend/action/rhnpackage/BaseSystemPackagesAction.java
@@ -63,13 +63,6 @@ public abstract class BaseSystemPackagesAction extends RhnAction {
 
         Set<String> sessionSet = SessionSetHelper.lookupAndBind(request, getDecl(sid));
 
-        //if its not submitted
-        // ==> this is the first visit to this page
-        // clear the 'dirty set'
-        if (!requestContext.isSubmitted()) {
-            sessionSet.clear();
-        }
-
         SessionSetHelper helper = new SessionSetHelper(request);
 
         if (request.getParameter("dispatch") != null) {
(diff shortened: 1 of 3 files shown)


xkollar:
I can not recommend accepting 8d5497e.

* Legibility vs. Readability.
* Changes are not only about changes of indentation/making code more readable.
* Are suggested changes aligned with code style? Things might have changed since I have worked on the project but I do not think so.

(Also, probably someone else should review this `:-)`.)

jiridostal:
Strongly concur. 8d5497e cannot be accepted as it is at the moment - you could try separating it into "code style" and "code change" commits. However, remaining two commits seem reasonable.

tlestach:
I'd question the motivation for this PR. Current behavior is expected and I'd say it corresponds with the original requirements.
Listed pages are not of that kind, where I'd like to preserve the selections.

In addition to that there's no response to reactions since Jul 15, 2017, so I'm rejecting the PR.

PR #554, opened 2017-07-03
rhnsd - 1383668 - close and reopen syslog when redirecting child output

mmatuska:
If rhnsd runs in daemon mode, stdin is redirected to /dev/null (if possible) and syslog uses fd 1 (stdout) for its socket. We should not close the syslog socket with a dup2() without a closelog(). When redirecting stdout and stderr from the rhn_check child process, fd of the syslogd socket may interfere. Make sure syslogd reopens on a safe fd in the child process.


jdobes:
It makes sense, thank you.

PR #555, opened 2017-07-06
Use ErrorStatusFilter in case of error and clean up UI messages

ncounter:
It can happen that after the `client request` is completed with success, something else in the back-end fails: this leads to an inconsistent page with some *success message* about the `request` and **at the same time** an `ISE` page content.

Inconsistent UI:
screenshot from 2017-06-22 09-48-16

This can be caused, for instance, in `/rhn/systems/customdata/CreateCustomKey.do` page using a too long sized string as a value for the `input` or `textarea`. Nothing will fail for the `client request` since adding a very long `String` in an `Hibernate Object` is not a problem. The `error` will be raised up as soon as [Hibernate commit a transaction](https://github.com/spacewalkproject/spacewalk/blob/master/java/code/src/com/redhat/rhn/frontend/servlets/SessionFilter.java#L58), the Database will fail `throwing` an `Exception` after the `response` is already completed, containing *success messages*.

The purpose of this PR is to
 - [activate the ErrorStatusFilter.java](https://github.com/ncounter/spacewalk/commit/8bf1aeb0daeea628918c85bdb808161b1fdf745f#diff-12d666402bb5c2b88d32e3d57ac18db6R154)
- [clean up success messages to prevent showing them in the UI](https://github.com/ncounter/spacewalk/commit/f49dda7fa7501e6c681e4dc73a8ee70b14f81dda#diff-8e5e591f11c105ac2bbf0466e1ba4d92R50)
- [add a custom generic message about the occurred exception](https://github.com/ncounter/spacewalk/commit/b10345cfeab524fdc5a31faef2ea0d3b9652621a#diff-bc8cde1cf0e011dde44e0965b7abf054R77)

UI after this patch:
screenshot from 2017-07-04 13-38-24

### DIFF
--- a/java/code/src/com/redhat/rhn/frontend/servlets/ErrorStatusFilter.java
+++ b/java/code/src/com/redhat/rhn/frontend/servlets/ErrorStatusFilter.java
@@ -14,6 +14,8 @@
  */
 package com.redhat.rhn.frontend.servlets;
 
+import org.apache.struts.Globals;
+
 import java.io.IOException;
 
 import javax.servlet.Filter;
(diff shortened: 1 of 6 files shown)


jiridostal:
I've successfully reproduced this issue and I think that I had bumped at the similar behavior before. I've reviewed the code and retested. Works as described.

PR #556, opened 2017-07-13
Added a script that performs post-processing on DocBook XML output.

adahms:
This PR adds a Python-based script that allows users to perform post processing on the apidoc-docbook output as follows:

* Adds a 'Name' attribute for all parameters and methods
* Adds bold formatting to all parameters marked up using quotation marks


MichaelMraka:
Thanks for your contribution.

PR #565, opened 2017-08-11
Enhanced documentation for Taskomatic memory variables in rhn.conf 

csheremeta:
Corrected variables for Taskomatic memory in rhn.conf  and added documentation for taskomatic.java.initmemory. It is important for users to use the correct variables in rhn.conf when changing these settings, as these values will persist through upgrades, whereas changes made in the config-default files will not persist through upgrades.

### DIFF
--- a/spacewalk/config/usr/share/man/man5/rhn.conf.5
+++ b/spacewalk/config/usr/share/man/man5/rhn.conf.5
@@ -215,11 +215,18 @@ then the job will be queued.
 2
 
 .TP
-.B "taskomatic.maxmemory" (integer)
+.B "taskomatic.java.maxmemory" (integer)
 The maximum amount of memory (MB) that Taskomatic can use. If you find that Taskomatic is running out of memory, consider increasing this.
 .IP
 .B Default:
 1024
+
+.TP
+.B "taskomatic.java.initmemory" (integer)
+The initial amount of memory (MB) that Taskomatic is allocated on start-up.
+.IP
+.B Default:
+256
 
 .TP
 .B "package_import_skip_changelog" (boolean)


ogajduse:
Adding link to [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1480810) to be tracked alongside with this PR.

spacewalk/config/usr/share/man/man5/rhn.conf.5
```diff
@@ -215,13 +215,20 @@ then the job will be queued.
 2
 
 .TP
-.B "taskomatic.maxmemory" (integer)
+.B "taskomatic.java.maxmemory" (integer)
```
csheremeta: Hey Tomas,

No, the taskomatic.java.[init/max]memory is correct, I think because rhn.conf does not know that wrapper.java.maxmemory would refer to Tasko's memory (just a guess). For actual proof:

[root@sat58mmello ~]# grep memory /etc/rhn/rhn.conf
taskomatic.java.initmemory=512
taskomatic.java.maxmemory=1192

[root@sat58mmello ~]# service taskomatic stop
Stopping RHN Taskomatic...
Stopped RHN Taskomatic.

[root@sat58mmello ~]# ps aux | grep memory | head -1
root      4285  0.0  0.0  19132   672 ?        Sl   18:06   0:00 /usr/bin/taskomaticd /usr/share/rhn/config-defaults/rhn_taskomatic_daemon.conf wrapper.java.initmemory=512 wrapper.java.maxmemory=1192 wrapper.pidfile=/var/run//taskomatic.pid wrapper.daemonize=TRUE

  ^ it worked

tkasparek: After applying following to rhn.conf
taskomatic.java.initmemory=512
taskomatic.java.maxmemory=1192

I can see following:
root     23361  0.0  0.0  21180   680 ?        Sl   09:34   0:00 /usr/bin/taskomaticd /usr/share/rhn/config-defaults/rhn_taskomatic_daemon.conf wrapper.maxmemory=2000 wrapper.java.initmemory=512 wrapper.java.maxmemory=1192 wrapper.pidfile=/var/run//taskomatic.pid wrapper.daemonize=TRUE

I guess having wrapper.maxmemory=2000 and wrapper.java.maxmemory=1192 together won't work well.

Putting following into rhn.conf (removed .java):
taskomatic.initmemory=512
taskomatic.maxmemory=1192

which results in:
root     23660  0.0  0.0  21176   680 ?        Sl   09:38   0:00 /usr/bin/taskomaticd /usr/share/rhn/config-defaults/rhn_taskomatic_daemon.conf wrapper.maxmemory=2000 wrapper.initmemory=512 wrapper.maxmemory=1192 wrapper.pidfile=/var/run//taskomatic.pid wrapper.daemonize=TRUE

where I'd say latter takes precedence so it should be OK.

Which is weird as in /usr/share/rhn/config-defaults/rhn_taskomatic_daemon.conf there's .java in config option name
wrapper.java.initmemory=256
wrapper.java.maxmemory=1024

I guess I need to investigate a bit more :)

PR #571, opened 2017-09-12
bz1490801 - [RFE] skip child channels with no repo during sync

optiz0r:
Prevent `spacewalk-repo-sync` from aborting if any child channel doesn't
have a repository/URL associated, by removing them from the list of
channels to be synchronised.

Allows syncing using `-p` option when some channels don't have upstream
repositories (i.e. they contain packages uploaded via rhnpush or webui).

Fixes bz1490801


jdobes:
Makes sense, thanks.

PR #575, opened 2017-09-22
1494575 - RHEL7.* isn't subscribed to EUS channel with option --use-e…

Pajinek:
The function rpm.labelCompare doesn't compare correct version of package. For example 7.1-1.el7_1' and '7.1-1.el7' it is not same version.


jdobes:
Travis CI build failed because colon is missing on line 860.

tkasparek:
Weird I've just registered 7.1 server to rhel-x86_64-server-7.1.eus on two different Satellites.

# cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 7.1 (Maipo)

tkasparek:
Fixed by more general approach patches.

PR #579, opened 2017-09-29
Find nearby version of eus channel to subscribe

Pajinek:
For example: 
release: redhat-release-server-7.0-1.el7.x86_64

channels:
rhel-x86_64-server-7.1.eus
rhel-x86_64-server-7.2.eus
rhel-x86_64-server-7.3.eus
rhel-x86_64-server-7.4.eus

The system will be registered to "rhel-x86_64-server-7.1.eus".

### DIFF
--- a/client/rhel/rhn-client-tools/src/bin/rhnreg_ks.py
+++ b/client/rhel/rhn-client-tools/src/bin/rhnreg_ks.py
@@ -133,6 +133,18 @@ class RegisterKsCli(rhncli.RhnCli):
                                                    self.options.password)
             other['channel'] = channels['default_channel']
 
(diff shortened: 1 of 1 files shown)


tkasparek:
Can this somehow be done on server instead of on client?

Pajinek:
Yes. I suppose that it is possible to implement it on server side. But it looks more that it is client issue which wasn't implemented on client side. Server side is prepared for it.

tkasparek:
Closing this PR for now as I'd like to investigate whether this can be done on client, and whether it actually makes sense to find the nearest EUS channel (I suppose we can do this by adding new switch in order not to change current behavior).

PR #580, opened 2017-09-29
bz1497216 - [RFE] Add softwarechannel_setsyncschedule --latest

optiz0r:
Add support for the `--latest` option on the
`softwarechannel_setsyncshedule` command so that tasko sync tasks
only download the most recent versions rather than the full repository
content.

This brings the spacecmd command closer into line with the features
available via the webinterface and allows use of the latest option when
setting sync schedules programmatically.

Fixes [#1497216](https://bugzilla.redhat.com/show_bug.cgi?id=1497216)


tkasparek:
Thank you!
I've extended your pull request to include all sync options, like no-errata in softwarechannel_setsyncschedule in c4e5afa27238659d49cb6a2df2338a1ffa0cc396

PR #582, opened 2017-10-02
Fix typo introduced in 79b0ed8e2f

rpasche:
Just fix a simple typo introduced in 79b0ed8e2fa but it breaks package creation and travis tests.


rpasche:
Has been merged with #583

PR #581, opened 2017-10-02
Do not fail import on ignored errata

rpasche:
When a errata is ignored, because a newer errata has already
been found, this breaks the import with error

ERROR: 'list' object has no attribute 'keys'

in '_fix_erratum_packages' because 'packages' of an ignored errata
has not been 'transferred' to a dict() by '_fix_erratum_packages_lookup'

PR #587, opened 2017-10-19
Do not import ignored errata

rpasche:
@tkasparek 

See this PR as alternative for #581 

It is further modified, to not only fix the `ERROR: 'list' object has no attribute 'keys'`, I'm also able to import erratas for SLES11.3 (and older)

I don't think, that this patch should cause any trouble with "new" (current) repositories with very large (and mostly unique) `advisory_names`, but solves my problem in still syncing "SLES11.3", which is still supported. Beginning with SLES11.4, the `advisory_names` got longer and more unique.

I also have to note, that I have to wait for new patches in SLES11.3 to see, if this patch really works as designed.

### DIFF
--- a/backend/server/importlib/errataImport.py
+++ b/backend/server/importlib/errataImport.py
@@ -18,6 +18,7 @@
 
 from spacewalk.common.rhnException import rhnFault
 from importLib import GenericPackageImport
+from spacewalk.satellite_tools.syncLib import log
 
 
 class ErrataImport(GenericPackageImport):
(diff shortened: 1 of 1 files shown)


tkasparek:
Thank you, I'll test tomorrow that this does not break syncing of Red Hat and Fedora content and if everything is OK I'll accept this PR.

I suppose #581 can be closed.

rpasche:
Yes. #581 can be closed in favour of this one.

tkasparek:
Hi, I've just confirmed that nothing breaks for syncing errata coming from EPEL 6 repo so I'd accept your PR.
Shall I accept it right now or do you want to wait for new pathces in SLES11.3 for you to confirm that everything works well on your end?

rpasche:
Hi, I would really like to wait till next week and see, if updates come in on SLES11.3. Till now, I was unable to see, what happens, if an already imported advisory gets `updated`. I'm not sure, if I saw some code for "updating" (or remove and insert) an existing errata.

rpasche:
Hi @tkasparek 

I further found a small bug. An advisory with for example release '8159' was greater (newer) than an advisory with release '10245'. So `release` gets temporarily converted to a `long()`.

If you could - again - check, if it still does not break errata import on CentOS and EPEL, you can accept this PR.

Although a "new" errata with the same `advisory_name` would NOT get imported, this should only happen on old repositories, like SLES11.3. And if this happens, the workaround would be, to remove the already imported errata and sync again. Then, only the new one would get imported again.

tkasparek:
I'll recheck this week and of everything is OK I will accept this PR as I understood that I ave greenlight from you.

rpasche:
@tkasparek You have greenlight :)

PR #588, opened 2017-10-19
show list of arches for channel

Pajinek:
Missing command which shows all usable arches for creating new channel.

### DIFF
--- a/spacecmd/src/lib/softwarechannel.py
+++ b/spacecmd/src/lib/softwarechannel.py
@@ -630,6 +630,26 @@ def do_softwarechannel_listerrata(self, args):
 
 ####################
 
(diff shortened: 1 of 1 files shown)


spacecmd/src/lib/softwarechannel.py
```diff
@@ -630,6 +630,26 @@ def do_softwarechannel_listerrata(self, args):
 
 ####################
 
+def help_softwarechannel_listarches(self):
+    print("softwarechannel_listarches: lists the potential software"
```
ogajduse: There is missing parenthesis on line 634.

Pajinek: thanks.

PR #589, opened 2017-10-19
move deletion to the inner loop to delete all duplicates

lucidd:
This moves the duplicate deletion to the inner loop so all duplicates are deleted not just one.


tkasparek:
Thanks for the catch!

PR #601, opened 2017-10-28
Python gzipstream for python3

mcalmer:
Make syntax python3 compatible and build a python3 subpackage


jdobes:
Merged, thanks.

When looking at this I also changed 00948342fc920dd5e34a9773b3e7419cd2146791, it seems as a bug.

PR #603, opened 2017-11-06
fix joining strings

Pajinek:
>> satellite-sync -m dump -c rhn-tools-rhel-x86_64-server-7
Downloading:+++ sending log as an email +++
SYNC ERROR: attempting to display as much information as possible:

(Check logs/email for potentially more detail)

TypeError("cannot concatenate 'str' and 'int' objects",)
cannot concatenate 'str' and 'int' objects


jdobes:
This is a regression caused by PR #587.

jdobes:
Reason why this was missed in PR #587 is probably because satellite-sync populates Errata objects to import with release as number while spacewalk-repo-sync with release as string.

PR #605, opened 2017-11-15
Add a command-line switch to satellite-sync to ignore proxy settings

lrochfor-oracle:
This patch adds a command-line switch to allow satellite-sync to perform ISS against an internal Spacewalk server when an HTTP proxy server has been configured in the environment or config file.


MichaelMraka:
Merged to master. Thanks.

PR #607, opened 2017-11-20
Overview page small fixes

renner:
## What does this PR do?

This patch fixes two small issues on the overview / landing page that you go to immediately after login.

## GUI diff

Screenshots were taken from SUSE Manager.

Before:
overview-page-before

After:
overview-page-after

## Previous Behavior

1. Wrong string was referenced for system groups: It was showing "System Group Name" instead of "System Groups". "System Group Name" is supposed to be just the title of the second column in that table.
2. The title of "Recently Registered Systems" was in `<strong>` tags which makes it look odd and different from the other section titles.

## New Behavior

1. Correct string is now used for the title of system groups: "System Groups".
2. The `<strong>` tags were removed so the title appears equally formatted as other section titles.


MichaelMraka:
Thanks.

PR #610, opened 2017-12-01
Display constraint messages more user friendly

clanig:
This pull request applies the following changes:
1. The message when you enter a key limit for custom keys is more comprehensive to end-users.
2. The `NumericConstraint` which was implemented with `Double`-values is split into a `DoubleConstraint` and a `LongConstraint` class. As a result the constraints for big whole numbers stay accurate until `Long.MAX_VALUE` and `Long.MIN_VALUE` instead of +/- 2^53. There is no numeric constraint without limits anymore. Both variables are initialized in the constraint constructor with `MIN_VALUE` and `MAX_VALUE` of `Double`/`Long`. An additional JUnit test has been added.
3. A big number in the database is not translated into an integer anymore. Adding a limit exceeding the range of `Integer` led to a negative key limit being displayed in the UI because of the overflow in the translation routine.
4. The method name `isValid` which is not returning a boolean has been renamed to `validate`, `TODO`-mark removed. An unused variable has been removed as well.

### DIFF
--- a/java/code/src/com/redhat/rhn/common/validator/DoubleConstraint.java
+++ b/java/code/src/com/redhat/rhn/common/validator/DoubleConstraint.java
@@ -25,10 +25,10 @@ import org.apache.log4j.Logger;
  * </p>
  * @version $Rev$
  */
-public class NumericConstraint extends RequiredIfConstraint {
+public class DoubleConstraint extends RequiredIfConstraint {
 
     /** Logger instance */
-    private static Logger log = Logger.getLogger(NumericConstraint.class);
+    private static Logger log = Logger.getLogger(DoubleConstraint.class);
 
     /** Minimum inclusive value allowed */
     private Double minInclusive;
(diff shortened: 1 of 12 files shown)


grinrag:
Hi Christian,
Thank you for the PR, I will take a look at it.

clanig:
Hi Longo,
thanks for taking a look at the PR. I have rebased the branch and updated one line in the DoubleConstraint constructor today because there was a typo. In particular the variable for the minimum value was initialized a second time with the max value of double.

jiridostal:
- [x] Change misleading name of method isValid to validate.
  - Makes sense
- [x] Display messages about wrong input more end-user friendly.

  - I've done some testing and didn't bump at any issue so far

- [x]  Add new JUnit test case for long constraint.

- [x] Remove unused local variable strValue
- [x] Change data type of key counters according to the database and constraint

I'm accepting this PR.

clanig:
Thank you for taking your time.

PR #614, opened 2017-12-24
Updated man page for rhn.conf file to include information on setting Satellite's SMTP server

csheremeta:
Added variable for SMTP server to rhn.conf file and updated the man page

### DIFF
--- a/spacewalk/config/usr/share/man/man5/rhn.conf.5
+++ b/spacewalk/config/usr/share/man/man5/rhn.conf.5
@@ -235,6 +235,13 @@ When importing packages, skip non-essential data that can make the import faster
 .B Default:
 0
 
+.TP
+.B "java.smtp_server"
+This parameter defines the server Satellite will use as its SMTP server.
+.IP
+.B Default:
+localhost
+
 .SS Database Options
 .TP
 .B "db_backend"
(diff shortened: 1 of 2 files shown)


spacewalk/config/usr/share/man/man5/rhn.conf.5
```diff
@@ -235,6 +235,13 @@ When importing packages, skip non-essential data that can make the import faster
 .B Default:
 0
 
+.TP
+.B "java.smtp_server"
+This parameter defines the server Satellite will use as its SMTP server.
+.IP
+.B Default:
+none
```
tchellomello: @csheremeta I think you should add the default as a localhost as we can see here on the code. 

```bash
↪ grep -R WEB_SMTP * 
common/conf/ConfigDefaults.java:    public static final String WEB_SMTP_SERVER = "java.smtp_server";
common/messaging/SmtpMail.java:        smtpHost = c.getString(ConfigDefaults.WEB_SMTP_SERVER, "localhost");
```

tlestach:
More info: https://bugzilla.redhat.com/show_bug.cgi?id=1528852

csheremeta:
updated Default in rhn.conf man page to localhost instead of none

csheremeta:
removed edits to rhn.conf file per comment from @tlestach on https://bugzilla.redhat.com/show_bug.cgi?id=1528852

PR #618, opened 2018-01-19
1536484 - Command spacecmd supports utf8 name of systems

Pajinek:
Fixed Bug 1536484. Name of system with utf8 chars causes error.

```
>> spacecmd system_search name:test
ERROR: 'ascii' codec can't encode characters in position 5-11: ordinal not in range(128)
```


tkasparek:
Wouldn't it be better to just remove all occurences of str() call in spacecmd?
git grep 'str(' | wc -l
61

Pajinek:
I checked it and I think that it is not necessary, because it is used for print helper or numbers. I prefere to create new pull request that focus on refactoring string's call in code:
```
str(len(failed)).rjust(3) -> "%3d" % len(failed)
str(action.get('id')).ljust(6) -> "%-6s % action.get('id')
```

tkasparek:
Closing due to conflicts, committed as 2ec2d415d9d309a614ed31ecb6ae1cf6f542b879

PR #626, opened 2018-02-16
Change man section 8 to 5

hayato-suzuki:
This file put in to man5 directory by make file.
Section 8 is wrong.

make file:
  "	install -m 644 rhnplugin.conf.5 $(PREFIX)/usr/share/man/man5/rhnplugin.conf.5"


tkasparek:
Section 8 seems more fitting to me than section 5 as per description of sections in `man man`

       5   File formats and conventions eg /etc/passwd
       8   System administration commands (usually only for root)

tkasparek:
Pardon me, I thought that PR fixes `man yum-rhn-plugin` while it fixes `man rhnplugin.conf`. In that case it makes sense to accept the PR.

PR #630, opened 2018-03-01
spacecmd-Added function to update software channel from spacecmd + Refactoring

admd:
## What does this PR do

This aim of this PR is to do two things

1.  Add a method in spacecmd so user an update the software channel from spacecmd.
2.  Clean some existing code


jiridostal:
Could you possibly split this into two commits? One for new a method and one for code cleanup

admd:
Hi @jiridostal , I have split the code into two commits. Let me know if you need any further changes.

admd:
@shastah you are correct.  Codebase I was working on was based on spacewalk 2.6 which didn't have a method for updating the details and then I just cherry-picked that commit to spacewalk2.7 without realizing that spacewalk 2.7 already have similar `softwarechannel_setdetails` method in spacecmd.  

This PR did more than this but I can remove spacecmd part in a new PR. @jiridostal ok?

jiridostal:
Sure, go for it. :+1:

PR #632, opened 2018-03-21
1365569 - Update SELinux to allow httpd to read systemid file

lrochfor-oracle:
SELinux prevents httpd from reading systemid file on a Spacewalk proxy. This causes client registration via the proxy to fail.

Update spacewalk-proxy-selinux policy to fix.

Please see BZ for details.


tkasparek:
Thank you Laurence

PR #633, opened 2018-03-26
Fix channel <-> advisory field inversion on the ErrataChannelIntersec…

angystardust:
…tion page
"A picture is worth a thousand words" :)
screenshot from 2018-03-26 13-10-05


jiridostal:
Thank you!

PR #637, opened 2018-04-06
Fix apt-transport-spacewalk to support signed repos

rpasche:
This is the *client* part of #636 to be able to use signed Debian / Ubuntu repositories. See https://bugzilla.redhat.com/show_bug.cgi?id=1198723 for more information.

It also looks like the package is going lost on Debian (see https://tracker.debian.org/pkg/apt-spacewalk). 

Therefore, I patched the files within this repository, so that we can be able to rebuild the package. I also increased the version of the Debian package to ```1.0.7-1```, so the package is newer than the last official debian package.

The creation of the Release and Release.gpg files still needs to be done manually on the server for each channel. See also @philicious blog entry http://www.devops-blog.net/spacewalk/gpg-signing-apt-repository-in-spacewalk

### DIFF
--- a/client/debian/packages-already-in-debian/apt-transport-spacewalk/50spacewalk
+++ b/client/debian/packages-already-in-debian/apt-transport-spacewalk/50spacewalk
@@ -6,10 +6,10 @@ APT {
   Update {
         List-Refresh "true";
         Pre-Invoke {
-            "if [ -x /usr/lib/apt-spacewalk/post_invoke.py ]; then /usr/lib/apt-spacewalk/post_invoke.py; fi";
+            "if [ -x /usr/lib/apt-spacewalk/pre_invoke.py ]; then /usr/lib/apt-spacewalk/pre_invoke.py; fi";
         }
   }
 };
 DPkg::Post-Invoke {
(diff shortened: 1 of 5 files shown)


niehsgst:
Might be worth pointing out that the blog entry is ~3 years old at this point, and in the interim Debian has deprecated SHA-1 signing of repositories.  For my part, this resulted in signed repositories that clients would nevertheless report as having 'invalid signatures', using @philicious' blog/instructions (see my comment on the bottom of the blog).

rpasche:
Am 10. April 2018 15:02:50 MESZ schrieb niehsgst <notifications@github.com>:
>Might be worth pointing out that the blog entry is ~3 years old at this
>point, and in the interim Debian has deprecated SHA-1 signing of
>repositories.  For my part, this resulted in signed repositories that
>clients would nevertheless report as having 'invalid signatures', using
>@philicious' blog/instructions (see my comment on the bottom of the
>blog).

Sorry, but I cannot confirm that.

Also, the script does not use SHA1 sum. So where did you use that?

Maybe the error is within your channel, that the checksum SHA1 is used in there.

rpasche

-- 
sent from my mobile device

PR #640, opened 2018-04-23
Fix race condition in lookup_evr and possibly similar others

moio:
## The problem

When a new package is added to the database (at client/minion registration time or during reposync), its EVR is inserted in `rhnPackageEvr` via the `lookup_evr()` stored procedure.

There is a small window of time in which the stored procedure can fail if multiple transaction try `INSERT`ing the same EVR at the same time. This has been successfully reproduced in Postgres.

The following traces appear in Postgres logs:

```
2018-04-18 09:51:51 CEST susemanager postgres ERROR:  duplicate key value violates unique constraint "rhn_pe_v_r_uq"
2018-04-18 09:51:51 CEST susemanager postgres DETAIL:  Key (version, release)=(1.12.5, 39.2) already exists.
2018-04-18 09:51:51 CEST susemanager postgres STATEMENT:  insert into rhnPackageEVR(id, epoch, version, release, evr) values (2483, NULL, '1.12.5', '39.2', evr_t(NULL, '1.12.5', '39.2'))
2018-04-18 09:51:51 CEST susemanager spacewalk ERROR:  query returned no rows
2018-04-18 09:51:51 CEST susemanager spacewalk CONTEXT:  PL/pgSQL function lookup_evr(character varying,character varying,character varying) line 25 at SQL statement
2018-04-18 09:51:51 CEST susemanager spacewalk STATEMENT:  select * from lookup_evr($1,$2, $3, $4) as result
2018-04-18 09:51:51 CEST susemanager spacewalk ERROR:  current transaction is aborted, commands ignored until end of transaction block
```

## The Reproducer

To reproduce, run the [hammer_evr.rb](https://gist.github.com/moio/9bdb684102b1af89db3f47ad8ab2bc62) script against a server in multiple parallel copies at the same time:

```sh
$ ruby hammer_evr.rb & ruby hammer_evr.rb &ruby hammer_evr.rb &ruby hammer_evr.rb &ruby hammer_evr.rb & ruby hammer_evr.rb & wait
```

Note: change connection details, if needed, on line 12.

The script will produce the following error:

```
hammer_evr.rb:25:in `exec_prepared': ERROR:  query returned no rows (PG::NoDataFound)
CONTEXT:  PL/pgSQL function lookup_evr(character varying,character varying,character varying) line 25 at SQL statement
        from hammer_evr.rb:25:in `block (2 levels) in <main>'
        from hammer_evr.rb:20:in `each'
        from hammer_evr.rb:20:in `block in <main>'
        from hammer_evr.rb:18:in `transaction'
        from hammer_evr.rb:18:in `<main>'
```

## The proposed solution

This PR changes `lookup_` functions to use a feature available starting with Postgres 9.5 (the `ON CONFLICT` clause) to avoid this race condition. This approach was briefly discussed in `#spacewalk-devel` and it is already implemented in SUSE Manager. So far, our testing does not indicate regressions but still this is a vital area of the product, so please review carefully.

If the approach is OKed I will also produce migration scripts.

### DIFF
--- a/schema/spacewalk/postgres/procs/create_pxt_session.sql
+++ b/schema/spacewalk/postgres/procs/create_pxt_session.sql
@@ -26,11 +26,15 @@ declare
 begin
     l_id := nextval( 'pxt_id_seq' );
 
-    perform pg_dblink_exec(
-        'insert into PXTSessions (id, value, expires, web_user_id) values (' ||
-        l_id || ', ' || coalesce(quote_literal(p_value), 'NULL') ||
-        ', ' || p_expires || ', ' || coalesce(quote_literal(p_web_user_id), 'NULL') || '); commit');
+    insert into PXTSessions (id, value, expires, web_user_id)
+        values (l_id, p_value, p_expires, p_web_user_id)
+        on conflict do nothing;
 
-	return l_id;
+    select id
+        into strict l_id
+        from PXTSessions
+        where value = p_value and expires = p_expires and web_user_id = p_web_user_id;
+
+	  return l_id;
 end;
 $$ language plpgsql;
--- a/schema/spacewalk/postgres/procs/lookup_checksum.sql
+++ b/schema/spacewalk/postgres/procs/lookup_checksum.sql
@@ -34,25 +34,25 @@ begin
 
     if not found then
         checksum_id := nextval('rhnchecksum_seq');
-        begin
-            perform pg_dblink_exec(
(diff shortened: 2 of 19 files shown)


tkasparek:
Hello,

thank you for bringing this up. However I am afraid that on CentOS 6 which is still supported there's only PostgreSQL in 8.4 version available.
I guess there's no way how to make this working also with lower version of PostgreSQL, am I right?

Thank you

PR #641, opened 2018-04-23
[selinux] Really fix pam authentication

angystardust:
I'm sorry but I've just noticed that my previous selinux patch was missing the "create" permission on the *netlink_audit_socket* for the *tomcat_t* labelled processes. This should definitely fix the following 2 bugs:

* [Bug 1517791](https://bugzilla.redhat.com/show_bug.cgi?id=1517791) - AVC: Kerberos user cannot login to Spacewalk on RHEL 7 
* [Bug 1494675](https://bugzilla.redhat.com/show_bug.cgi?id=1494675) - PAM authentication no longer works after upgrading to CentOS 7.4


angystardust:
As for #642, it would be very useful to backport this PR to the 2.8 branch.

jiridostal:
Thanks again, this issue has been there for some time if I remember correctly. We don't backport patches into released version until it's security or major issue (#642). However as it's completion of partial fix, we could do that.

PR #645, opened 2018-04-27
SW 2.8: Fix KeyError: 'authtoken' when running spacewalk-repo-sync with --type deb

cbajumpaa:


backend/satellite_tools/repo_plugins/deb_src.py
```diff
@@ -25,6 +25,13 @@
 from spacewalk.satellite_tools.repo_plugins import ContentPackage, CACHE_DIR
 from spacewalk.satellite_tools.syncLib import log2
 from spacewalk.common.rhnConfig import CFG, initCFG
+from urlgrabber.grabber import URLGrabError
```
tkasparek: This import is not needed, could it be removed please?

PR #651, opened 2018-05-15
Fix crash in satellite_tools on RH/Fedora/Mageia urlgrabber

Conan-Kudo:
The `authtoken` parameter isn't available in non-SUSE `urlgrabber`, which causes it to crash when it attempts to execute this code. Checking for the existence of of the key in the `params` dictionary prevents the failure while preserving compatibility with SUSE systems.


Conan-Kudo:
Can this also be backported to Spacewalk 2.8? This issue caused the repo sync functionality to fail on my servers...

tkasparek:
Thank you for reporting this issue and for the patch.
I've backported it into 2.8 as well.

PR #653, opened 2018-05-28
client/rhel: Enable DNF plugin for Mageia 6+ and openSUSE Leap 15.0+

Conan-Kudo:
Since Mageia 6 and openSUSE Leap 15.0, DNF has been available in the distributions for usage. In addition, Mageia 6 offers DNF as a fully supported package manager option, with plans to transition to it as the only system package manager in the next release.

This will permit Mageia and openSUSE machines using DNF to work with Spacewalk.

In addition, the yum plugin was trivially adapted to be able to build for Mageia, too, since Mageia has a working yum now, too.

### DIFF
--- a/client/rhel/rhn-client-tools/rhn-client-tools.spec
+++ b/client/rhel/rhn-client-tools/rhn-client-tools.spec
@@ -1,4 +1,4 @@
-%if 0%{?fedora} || 0%{?suse_version} > 1320 || 0%{?rhel} >= 8
+%if 0%{?fedora} || 0%{?suse_version} > 1320 || 0%{?rhel} >= 8 || 0%{?mageia}
 %global build_py3   1
 %global default_py3 1
 %endif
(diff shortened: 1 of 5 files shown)


Conan-Kudo:
The Travis log indicates something is broken with the tito setup for `rhn-client-tools`, and I don't really know how to fix that...

client/rhel/rhn-client-tools/rhn-client-tools.spec
```diff
@@ -256,11 +270,17 @@ Requires: rhn-setup-gnome = %{version}-%{release}
 Requires: python-gnome python-gtk
 %else
 Requires: pygtk2 pygtk2-libglade
+%if 0%{?mageia}
+Requires: usermode
+%else
```
tkasparek: This %if %else condition is missing the %endif section which causes the travis build to fail.

tkasparek:
Hello, see my comment in the source, there's an if/else statement which is missing the endif, section thus "rpm -q --specfile rhn-client-tools.spec" fails with:
error: line 264: Unclosed %if
error: query of specfile rhn-client-tools.spec failed, can't parse

Could you please fix it by adding missing %endif? (I could do that for you, but I am sure you know better where it should belong). The rest of the specfiles looks good so this is the only issue blocking this PR.

client/rhel/rhn-client-tools/rhn-client-tools.spec
Conan-Kudo: Should be fixed now.

Conan-Kudo:
@tkasparek Looks like it succeeded this time. 🥇

PR #656, opened 2018-06-12
client/debian: Port apt-spacewalk to be Python 3 ready

Conan-Kudo:
All of apt-spacewalk's dependencies are Python 3 ready in the latest Debian and Ubuntu releases, but this is not necessarily true for earlier releases that may be actively managed by Spacewalk today.

The dsc packaging for apt-spacewalk is adjusted so that `python-six` is a runtime dependency of apt-spacewalk, since the code uses `six` for `urlparse()` now.


MichaelMraka:
Thank you for fixing debian packages!

Conan-Kudo:
@MichaelMraka This is only the beginning!

PR #657, opened 2018-06-26
Fix system-currency report

redflo:
SLES errata have lower case synopsis: "important: ..." and not "Important: ...".


tkasparek:
Hello @redflo JFYI similar queries can be find at multiple different places, e.g. at java/code/src/com/redhat/rhn/common/db/datasource/xml/System_queries.xml which I've just fixed by 6e95d151b734715ca760c39e00b66b94ce6999b1

PR #661, opened 2018-07-11
Add a debug flag to the spacewalk-repo-sync utility

EdgeJ:
Allows spacewalk-repo-sync to print messages to stdout instead of logging to a file, for simplifying debugging sync issues.


tkasparek:
Hello,
thank you for your first time contribution :-)

JFYI I've seen following error in the output with your changes:
Please check 'reposync/rhel-x86_64-server-7.log' for sync log of this channel.
chgrp: cannot access `stdout': No such file or directory

I've fixed this issue for you in the 767bfe9f966f8666663f9f916fb4cb25f6510d80

EdgeJ:
Thanks for the fixup! Don't have SuSE, so I hadn't tested that block.

--John

tkasparek:
Actually I got the error on RHEL 6 system as the chgrp is executed also on non SuSE systems, the only difference there that there's group www for SuSE instead of apache.

PR #666, opened 2018-07-31
Fix: errata IDs must be unique

mbologna:
Errata IDs must be unique when cloning channel erratas. This patch
make sure that errata ID will not be duplicated before inserting them
into the database.

Fix for:
* [Bug 1539095](https://bugzilla.redhat.com/show_bug.cgi?id=1539095) - spacecmd softwarechannel_clonetree aborts with duplicate key value violates unique constraint "rhn_ce_ce_uq"


tkasparek:
Thank you!

PR #667, opened 2018-08-21
Fix icon typo

ncounter:
A probably *never-used* icon (I couldn't find any usage of `header-errata-set-add`), but at least it needs a typo fix - as per [its CSS rule definition](https://github.com/spacewalkproject/spacewalk/blob/1c236c59cc87e9dcc7f6251dbf19926a83fad033/branding/fonts/font-spacewalk/css/spacewalk-font.css#L62) - if it will be ever used in the future.


tkasparek:
Thank you!

PR #669, opened 2018-09-25
Fixes proxy support for deb repos

pandiloko:
I never managed to synchronize deb repos behind a proxy. It actually works from command line if http_proxy variable is set because, as stated in the docs, Python request honors env vars. I'm not an experienced Python dev but the proposed changes seemed to work for me.

### DIFF
--- a/backend/satellite_tools/repo_plugins/deb_src.py
+++ b/backend/satellite_tools/repo_plugins/deb_src.py
@@ -62,7 +62,7 @@ class DebPackage(object):
 
 class DebRepo(object):
     # url example - http://ftp.debian.org/debian/dists/jessie/main/binary-amd64/
-    def __init__(self, url, cache_dir, pkg_dir, proxy_a="", proxy_u="", proxy_p=""):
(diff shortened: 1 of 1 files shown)


backend/satellite_tools/repo_plugins/deb_src.py
```diff
@@ -62,7 +62,7 @@ def is_populated(self):
 
 class DebRepo(object):
     # url example - http://ftp.debian.org/debian/dists/jessie/main/binary-amd64/
-    def __init__(self, url, cache_dir, pkg_dir):
+    def __init__(self, url, cache_dir, pkg_dir, proxy_a="", proxy_u="", proxy_p=""):
```
Conan-Kudo: Please use more meaningful names for the parameters of this function.

For example: `proxy_addr`, `proxy_user`, and `proxy_pass` are better names here?

pandiloko: |I've done the suggested changes

PR #672, opened 2018-10-05
Fix wrong "python3-newt" dependency name for SUSE systems

meaksh:
This PR fixes a wrong `python3-newt` dependency name for `rhn-client-tools` on SUSE systems.


meaksh:
@Conan-Kudo - Thanks.

Yes, the `python3-newt` package is available for both Leap 15.0 and Tumbleweed via the official channels:

https://build.opensuse.org/package/show/openSUSE:Leap:15.0/newt
https://build.opensuse.org/package/show/openSUSE:Factory/newt

PR #677, opened 2018-12-05
Fix typo lablel

pipopopo:


utils/spacewalk-manage-channel-lifecycle
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python2
+#!/usr/bin/python
```
Conan-Kudo: This will break things. This was intentionally changed since no one has ported it to Python 3.

pipopopo: True, only changes I wanted to make are:
[https://github.com/spacewalkproject/spacewalk/pull/678/files](https://github.com/spacewalkproject/spacewalk/pull/678/files)

pipopopo:
Dunno what happened here but seems like there are more changes then I made maybe because of 
pipopopo wants to merge 2 commits into spacewalkproject:master from "unknown repository"

PR #679, opened 2018-12-19
clean-up of Config singleton to allow new config path and add proper handling of http proxies

vlad-belogrudov:


MichaelMraka:
Is the purpose of this PR to fix tests? Because I can see only changes in tests no new ability added.

PR #680, opened 2019-01-28
Spacewalk 2.9 Systems List IP Address

taylorstrange:
Hello, 

I'd like to be able to see a system's IP address from the system list.  Is there a way to add/remove columns from that system list? 

The furthest I've gotten in trying to achieve this effect is by modifying the "rhnserveroverview" view in the database. But's it's quite beyond me on how to see that change reflected in the web interface.


MichaelMraka:
> Hello,
> I'd like to be able to see a system's IP address from the system list. Is there a way to add/remove columns from that system list?

Which system list? On which page?
Generally list views in spacewalk are not customizable.

> The furthest I've gotten in trying to achieve this effect is by modifying the "rhnserveroverview" view in the database. But's it's quite beyond me on how to see that change reflected in the web interface.

To modify a page you need to modify corresponding .jsp and most likely also its java class.

taylorstrange:
The page when you click on "Systems".  rhn/systems/SystemList.do

Thanks for your help.

MichaelMraka:
Hello, this is not a good place to file RFEs or bugs.
See how to file a bug: https://github.com/spacewalkproject/spacewalk/wiki#want-to-file-a-bugrfe

PR #685, opened 2019-02-07
Spacewalk Web UI accept connections only after redirect from Okta.

taylorstrange:
Okta simple web authentication (SWA) works nicely with spacewalk especially without having to set up and IPA domain.  I still have to provision users in spacewalk but I can use PAM to leverage the user's AD credentials.

I would really like to be able to leverage Okta's multi-factor authentication (MFA) capability with SWA to Spacewalk.  Right now I have that enabled. However users can still bypass Okta if they go directly to the spacewalk site.  Can I configure spacewalk or httpd to accept connections that have only been redirected from Okta? In other words, only connections with a specific string in the headers of the request?


Conan-Kudo:
@taylorstrange This is not the right place to file RFEs. Please use Bugzilla as directed on the website.

MichaelMraka:
Hello, this is not a good place to file RFEs or bugs.
See how to file a bug: https://github.com/spacewalkproject/spacewalk/wiki#want-to-file-a-bugrfe

PR #686, opened 2019-02-11
Extra semicolon leads to "SQL command not properly ended" with Oracle

DocFraggle:
See discussion at https://www.redhat.com/archives/spacewalk-list/2019-February/msg00000.html for details.

I couldn't test it, but as this is the only SQL statement ending with a semicolon, it has to be the reason for ORA-00933: SQL command not properly ended

Regards, Christian


DocFraggle:
OK, I was able to test the change.
I extracted the corresponding JAR file, /usr/share/rhn/lib/rhn.jar, removed the semicolon, recreated the archive and restarted Spacewalk. Afterwards, I was able to manage multiple systems without the internal server error.

Christian

MichaelMraka:
Thank you for the bugfix.

PR #688, opened 2019-04-07
Added resumable downloads support for apt transport

pterix:
Patch adds support of resuming downloading of packages for apt spacewalk clients. Usable if previous attempt of update was aborted or connection was lost.


MichaelMraka:

PR #689, opened 2019-04-16
Added SNI support for clients

pterix:
We need SNI support for our spacewalk clients. Spacewalk server work behind
the nginx web server and is one of a lot of virtual vhosts. For some reasons nginx uses
separate certificates for each virtual host, not one cert with many subjectAltName.
We have a problem with client connections in such environment.

Proposed changes use a call to set_tlsext_host_name() from SSLSocket when setting up
the SSL client connection, which allows connections to SNI-enabled servers. Tested on
many clients (under Ubuntu), work fine for us.


MichaelMraka:

PR #690, opened 2019-04-25
1703076 - Add search server JDBC idle time config options to rhn_sear…

lrochfort:
…ch.conf

Add C3P0 config options for maxIdleTime and
maxIdletimeExcessConnections to the search server mybatis config.

The search server uses mybatis for it's ORM persistence layer, which
in turn uses C3P0 for JDBC connection pooling. C3P0 is actually
configured without pooling, and with unlimited connection idle time by
default. This causes ORA-02396 errors when the target Oracle DB has
the IDLE_TIME property set in the user's profile because the
connection idle limit is exceeded.

Adding config options allows the search server's JDBC idle time to be
set so that it doesn't conflict with the DB's idle time.

The config options are set with the default of 0, meaning
unlimited. They may be overridden from the central /etc/rhn/rhn.conf file.

Orabug: 28537551

Reviewed-by: Alex Burmashev <alexander.burmashev@oracle.com>


MichaelMraka:

PR #691, opened 2019-05-09
Correct issue with dists in url hostname

jmccanta:
URLs like http://dists.examle.com/local/dists/ourmirror/ubunut/dists/... failed syncing.  The URL path was spit in the wrong place.  Patch to look for the rightmost /dists/ in the URL.  For standard debian naming, this should be fine.


MichaelMraka:
Thanks for the contribution.

PR #692, opened 2019-05-26
Add --json to format output as json not ruby string

jmccanta:
spacewalk-api outputs the results from xmlrpc calls as ruby strings.  Addining a --json option that outputs the results in json suitable to parse via other tools (e.g., jq)


jmccanta:
Issue with serializing dates.  withdrawing pull request

PR #694, opened 2019-05-29
Increase repo source url max lenght in frontend

kim-sondrup:
Increase repository source url max character lenght in frontend from 512 to 2048 so it match the DB schema changes that was made back in 2014.

Commit that changed the DB schema: a569b5e4cc06aad012f99419daa568cba3837db7


MichaelMraka:

PR #696, opened 2019-06-12
Added a systemd template for the rhnsd timer

satcom886:
Attempt to fix [this](https://bugzilla.redhat.com/show_bug.cgi?id=1688278) bug. As I understand it, rhnsd.service should be disabled and rhnsd.timer enabled.


MichaelMraka:
Thanks for the contribution.

PR #699, opened 2019-07-01
return a non-zero exit status on errors in rhn_check

zed:
Propagate exit status from the `rhn_check` child process.


MichaelMraka:
Fixed in 955b2a708ad46b87e8c150a054a07c95ca6e0ef8.
Thanks for the contributions.

PR #702, opened 2020-02-08
added treeinfo file to download list

sbluhm:
.treeinfo is not downloaded as it is hidden. This fix ensures its download. Missing treeinfo file causes errors during provisioning. virt-install would not detect the install source without this file.


sbluhm:
Seems to be a build system issue:
DEBUG util.py:583:  BUILDSTDERR: ERROR: not a btrfs filesystem: /var/lib/mock
DEBUG util.py:583:  BUILDSTDERR: ERROR: can't access '/var/lib/mock'
DEBUG util.py:732:  Child return code was: 1
DEBUG util.py:1505:  Please ignore the error above above about btrfs.

PR #704, opened 2020-03-11
Updated "Required" for clarity.

sbluhm:
Updated "Required" to clearly show what package provides the requirement.


MichaelMraka:
Neither certwatch or crypto-utils is available on RHEL/CentOS 8 (and even Fedora 31).
I've disabled this package for mentioned buildroots and closing this PR.

sbluhm:
This causes an issue with installing Spacewalk then. I added crypto-utils to my personal repository. I guess there are 3 options:
1. remove the dependencies to certwatch during setup and miss out on the functionality.
2. add crypto-utils to the side repositories.
3. move to a different provider with same functionality (something I would look at at a later stage).
Until this is solved, it would make it easier to identify the installation issues with this fix.

PR #711, opened 2020-03-13
Updated specs to build

sbluhm:
1. Updated a few spec files with sources and adapted the prep section if required.
2. Added source packages for redstone-xml and quartz from previous src.rpms on copr.


MichaelMraka:
Git is not a proper place for storing archives. We should come up with a different solution.

sbluhm:
ok, I will remove them. At least it won't make it worse.

sbluhm:
Is it OK like that or do you want me to squash this as well?

PR #714, opened 2020-03-15
spacewalk-java updates for Tomcat 9

sbluhm:
1. Updated spec file to also consider RHEL8 for Tomcat version check.
2. Added separate config template file for Tomcat 9 (path would give a warning in the Catalina logs otherwise)


MichaelMraka:
Could you please squash these commits into one? (e.g. without fixes and fixes of fixes).

sbluhm:
Will do. I think it will be two commits minimum: Update of spec and adding of new file.

sbluhm:
done

PR #721, opened 2020-05-04
Replaced yum with dnf

sbluhm:
Replaced yum with dnf


sbluhm:
Closing for time being. Just found an unresolved dependency. Will re-open once identified.

